PR #1, opened 2015-04-07
Initial PR to create Perl image 

jhadvig:


jhadvig:
@mnagy @bparees @mfojtik @soltysh PTAL

5.16/test/sample-test-app/index.pl
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/perl
+
+use Weather::Underground;
+
+$place = "Brno, Czech Republic";
+$weather = Weather::Underground->new(
```
bparees: i have mixed feelings about a test app that's dependent on a remote service.  i guess it's ok for now but if it starts flaking on us, we're going to want to switch to something self-contained.

bparees:
looks good @mnagy and @jhadvig !

5.16/test/sample-test-app/index.pl
jhadvig: @bparees I had mixed feeling myself but, realized that this test-app will serve just for the CI process(installing modules from cpanfile) for now. But I can see your point and will try to come up with some suitable example.

5.16/.sti/bin/assemble
```diff
@@ -0,0 +1,34 @@
...
+
+set -e
+
+echo "---> Installing application source"
+cp -Rf /tmp/src/* ./
+
+export CPAN_MIRROR=${CPAN_MIRROR:-""}
+
+# Don't test installed Perl modules by default
+if [ $ENABLE_CPAN_TEST = true ];then
```
mfojtik:
```
ENABLE_CPAN_TEST="--notest"
[ $ENABLE_CPAN_TEST != true ] && ENABLE_CPAN_TEST=""
export ENABLE_CPAN_TEST
```

5.16/.sti/bin/assemble
```diff
@@ -0,0 +1,34 @@
...
+	export ENABLE_CPAN_TEST=""
+else
+	export ENABLE_CPAN_TEST="--notest"
+fi
+
+# Installing dependencies with cpanfile
+if [ -f "cpanfile" ]
```
mfojtik: keep the `if cond; then` consistent

5.16/.sti/bin/assemble
```diff
@@ -0,0 +1,34 @@
...
+then
+	CPANM=`which cpanm`
```
mfojtik: isn't cpanm always installed in same location in centos7/rhel7 ? if so, then instead of doing this ugly `which` just check if the file exists

5.16/.sti/bin/assemble
soltysh: Shouldn't that be `if [[ "${ENABLE_CPAN_TEST-}" == "true" ]];`

5.16/.sti/bin/assemble
```diff
@@ -0,0 +1,34 @@
...
+	echo "---> Installing modules from cpanfile"
```
mfojtik: it is 'cpanfile' or 'Cpanfile'? Also please add ' ...' into **all** informative messages that reports progress of something or indicate that users will wait till something finish.

5.16/.sti/bin/run
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# For SCL enablement
+source .bashrc
+
+set -e
+
+exec httpd -C 'Include /opt/openshift/httpd.conf' -D FOREGROUND
+
```
mfojtik: remove empty line here.

5.16/.sti/bin/run
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# For SCL enablement
+source .bashrc
+
+set -e
```
mfojtik: why is -e needed here?

5.16/.sti/bin/usage
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+#
+# Default STI usage script for the perl-5.16 image.
+#
+DISTRO=`cat /etc/*-release | grep ^ID= | grep -Po '".*?"' | tr -d '"'`
+
+cat <<EOF
+This is a STI perl-5.16 ${DISTRO} base image:
```
mfojtik: I really wish we can copy this file from one version to another (and this applies to all other images). If there is a way to detect the current version, we should do it in bash way.

5.16/.sti/bin/assemble
jhadvig: What ? I dont understand. First you set the `ENABLE_CPAN_TEST` envvar to "--notest" and then you test it i fits not true ? What kind of logic is that ?

5.16/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+RUN yum install -y \
+	https://www.softwarecollections.org/en/scls/rhscl/httpd24/epel-7-x86_64/download/rhscl-httpd24-epel-7-x86_64.noarch.rpm \
+	https://www.softwarecollections.org/en/scls/rhscl/perl516/epel-7-x86_64/download/rhscl-perl516-epel-7-x86_64.noarch.rpm && \
+	yum install -y --setopt=tsflags=nodocs httpd24 perl516 perl516-mod_perl perl516-perl-CPANPLUS && \
+	yum clean all
+
+ENV PATH $PATH:/opt/rh/perl516/root/usr/local/bin
+
+# TODO: s save system?
```
mfojtik: save the world!

5.16/.sti/bin/usage
```diff
@@ -0,0 +1,17 @@
...
+To use it, install STI: https://github.com/openshift/source-to-image
+
+Sample invocation:
+
+sti build https://github.com/openshift/sti-perl.git --contextDir=/5.16/test/perl-test-app/ openshift/perl-516-${DISTRO}7 perl-sample-app
```
soltysh: We've decided not to have that beginning slash in `contextDir`, so it should be: 
`--contextDir=5.16/test/perl-test-app/`

5.16/Dockerfile.rhel7
```diff
@@ -0,0 +1,31 @@
+FROM openshift/base-rhel7
```
mfojtik: Sometimes I hate to see that we just copy && pasting Dockerfile content from centos7 to RHEL. It is easy to miss some important lines and it require that folks need to keep it mind that what you did in centos7 you have to backport here... Maybe we should consider using some Dockerfile.tpl file.

5.16/contrib/httpd.conf
```diff
@@ -0,0 +1,18 @@
+LoadModule perl_module modules/mod_perl.so
+DirectoryIndex index.pl
+
+#ServerRoot "/opt/openshift"
```
mfojtik: If there is a reason why is this commented, please provide the reason in appropriate comment.

5.16/test/run
```diff
@@ -0,0 +1,159 @@
...
+#
+# The 'run' performs a simple test that verifies that STI image.
+# The main focus here is to excersise the STI scripts.
+#
+# IMAGE_NAME specifies a name of the candidate image used for testing.
+# The image has to be available before this script is executed.
+#
+IMAGE_NAME=${IMAGE_NAME-openshift/perl-516-centos-candidate}
+
+declare -a WEB_SERVERS=(sample)
```
mfojtik: why using declare here?

5.16/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+# TODO: Cleanup cpanp cache?
+# TODO: We'll need to distribute these....
+RUN scl enable perl516 -- cpanp 's conf prereqs 1; s save system' && \
+	scl enable perl516 -- cpanp 's conf allow_build_interactivity 0; s save' && \
+	scl enable perl516 -- cpanp 'i App::cpanminus'
+
+# Since there are couple of issues with running httpd as a non-root user,
```
soltysh: Put a TODO here.

5.16/Dockerfile.rhel7
```diff
@@ -0,0 +1,31 @@
...
+ENV PATH $PATH:/opt/rh/perl516/root/usr/local/bin
+
+# TODO: s save system?
+# TODO: Cleanup cpanp cache?
+# TODO: We'll need to distribute these....
+RUN scl enable perl516 -- cpanp 's conf prereqs 1; s save system' && \
+	scl enable perl516 -- cpanp 's conf allow_build_interactivity 0; s save' && \
+	scl enable perl516 -- cpanp 'i App::cpanminus'
+
+# Since there are couple of issues with running httpd as a non-root user,
```
soltysh: Same here: TODO

5.16/test/sample-test-app/index.pl
mfojtik: +1 to Ben. We should not require external access from tests. This might break us in CI. The test app should be super simple and should just require some simple module you installed to verify that cpan is working properly.

5.16/test/run
soltysh: If you have just one, nuke it.

5.16/.sti/bin/assemble
mfojtik: @jhadvig point is to get rid of extra 'else'.

5.16/test/run
```diff
@@ -0,0 +1,159 @@
...
+      fi
+      break
+    fi
+    attempt=$(( $attempt + 1 ))
+    sleep $sleep_time
+  done
+  return $result
+}
+
+for server in ${WEB_SERVERS[@]}; do
```
soltysh: You don't have more than one, nuke this for loop.

5.16/test/sample-test-app/index.pl
soltysh: How about just simple hello world, with printing request vars or env vars like we do in all other images?

mfojtik: @soltysh i'm fine with hello world as long as it use something installed by assemble script (to verify that assemble do its job properly.)

README.md
```diff
@@ -1 +1,193 @@
...
+CentOS versions currently supported are:
+* CentOS7
+
+
+Installation
+---------------
+To build Perl image, choose between CentOS or RHEL based image:
+*  **RHEL based image**
+
+    To build a rhel-based perl-5.16 image, you need to run the build on properly
```
soltysh: on **a** properly

I'll be fixing that in other images.

5.16/.sti/bin/assemble
mnagy: It is 'cpanfile'.

5.16/.sti/bin/usage
soltysh: Besides the contextDir should point to an existing application, which currently should be:
`--contextDir=5.16/test/sample-test-app/`

5.16/.sti/bin/assemble
jhadvig: It's `cpanfile`. 
Also I dont see any point in adding ` ...` to the info messages. We dont do it other images - stin/ non-sti

README.md
```diff
@@ -1 +1,193 @@
...
+Test
+---------------------
+This repository also provides [STI](https://github.com/openshift/source-to-image) test framework,
+which launches tests to check functionality of a simple perl application built on top of sti-perl image.
+
+User can choose between testing perl test application based on RHEL or CentOS image.
+
+*  **RHEL based image**
+
+    To test a rhel7-based perl-5.16 image, you need to run the test on properly
```
soltysh: on **a** properly

5.16/.sti/bin/usage
jhadvig: Yep, my mystake I've changed the naming of the app during the the dev. process

README.md
```diff
@@ -1 +1,193 @@
...
+* **`<perl-version>`**
+
+    * **Dockerfile**
+
+        CentOS based Dockerfile.
+
+    * **Dockerfile.rhel7**
+
+        RHEL based Dockerfile. In order to perform build or test actions on this
+        Dockerfile you need to run the action on properly subscribed RHEL machine.
```
soltysh: on **a** properly

5.16/Dockerfile
jhadvig: I was asking @mnagy if we can get rid of this TODO's, but he insist that we keep them here.

mfojtik: it does not make any sense to me :-) what exactly you are trying to say here?

README.md
```diff
@@ -1 +1,193 @@
...
+Environment variables
+---------------------
+
+To set these environment variables, you can place them into `.sti/environment`
+file inside your source code repository.
+
+* **ENABLE_CPAN_TEST**
+
+    Will install all the cpan packages and run their tests. Default value is set to 
+    `false`
```
soltysh: Missing dot at the end of statement.

5.16/Dockerfile.rhel7
jhadvig: I dont see you point whats wrong here? The base image is set appropriately to base-rhel7 since this is rhel based Dockefile. Whats your point ?

5.16/.sti/bin/assemble
mfojtik: we should do it :-) it is standard way how to report progress in OpenShift v3, we should keep our-selfs consistent in this... We can update other images as well as follow up.

README.md
```diff
@@ -1 +1,193 @@
...
+
+* **CPAN_MIRROR**
+
+    This variable specifies mirror URL which will cpanminus use to install dependencies.
```
soltysh: Either:
mirror URL which cpanminus will use to install ...
or
mirror URL which will be used by cpanminux to install...

5.16/test/run
jhadvig: +1

hack/build.sh
```diff
@@ -0,0 +1,55 @@
...
+	fi
+
+	docker build -t ${TAG} . && trap - ERR
+}
+
+if [ -z ${VERSION} ]; then
+	# Build all versions
+	dirs=${VERSIONS}
+else
+	# Build only specified version on Perl image
```
soltysh: version **of** Perl

5.16/Dockerfile.rhel7
mfojtik: @jhadvig my point is that 90% of this file is a copy & paste from centos7 Dockerfile (with the only 2 differences 1) FROM and 2) is subscription manager). We should have a template (in future) and share as much as possible between these two to avoid syncing.

5.16/test/sample-test-app/index.pl
jhadvig: @mfojtik yep that was my point

README.md
jhadvig: :+1:

5.16/test/sample-test-app/index.pl
soltysh: Yes, and you can install some basic simple package just to verify it works.

5.16/Dockerfile.rhel7
soltysh: Big fat +1 on that idea!

5.16/Dockerfile
soltysh: +1 either make this todo more verbose or nuke it. The same applies to the next two.

5.16/.sti/bin/assemble
jhadvig: @mfojtik but it this way the `ENABLE_CPAN_TEST` will always return empty string, no matter with what value the user will run the container. This is not appropriate logic for this use-case. We need to just check the value, if it is `true` or if it's anything else, `false` or empty string.

5.16/.sti/bin/usage
soltysh: How about substituting that during build? Or even better we could print here `IMAGE_DESCRIPTION` env var.

5.16/.sti/bin/assemble
jhadvig: @mnagy could you pls react on this comment since you are the one responsible for this ? Thanks

5.16/.sti/bin/assemble
soltysh: +1

5.16/.sti/bin/assemble
soltysh: +1

5.16/.sti/bin/assemble
jhadvig: @mfojtik if it is the way we do it in OSv3, then I'm also in favor of changing it !!! :goberserk:

5.16/.sti/bin/assemble
mfojtik: how about --notest as default without possibility to change it? do you want to test all modules installed by default in production? if not, then nuke this if/else and just set it.

jhadvig: @mfojtik I think we should keep it as an option. Like it doesn't matter if it's in production imo. We also provide this functionality in v2, I don't see why shouldn't we do it also here.
if you check the logicx, you will see that the `--notest` is the default value, only if user sets the envvar to true, only then the installation process will run with tests.

5.16/Dockerfile.rhel7
jhadvig: Yes it doest make sense but I believe that we won't applied that logic in this image for now. Would be good to create a card for it.

soltysh: Of course, I just trelloed that: https://trello.com/c/E97t4ifD/543-template-mechanism-for-dockerfiles

5.16/.sti/bin/assemble
soltysh: @jhadvig yes it is planned to be in OSv3. So we should apply it already here as well.

5.16/.sti/bin/assemble
mnagy: @mfojtik it depends on the collection version, I'd keep it as it is, it will be easier to port the script.

5.16/.sti/bin/run
jhadvig: @mfojtik was thinking that the run script will probably grow and we will need to add it there anyway, so why dont put it there now? We got it in all other `run` scripts

mnagy: +1, it is not as far as I can tell.

5.16/.sti/bin/usage
mnagy: I don't like that myself. I vote for sourcing /etc/os-release if it exists and using the variables directly.

http://www.freedesktop.org/software/systemd/man/os-release.html

jhadvig: @mnagy we are speaking about lang/db name and version. You cant get the it from the /etc/os-release.
I would go with @soltysh suggestion. Comments ?

5.16/Dockerfile
mnagy: Please keep the todo's. We (or I) will sort them out later together with SCL perl maintainers.

mfojtik: @mnagy make them more verbose to us :-) that is all i'm asking for.

5.16/test/sample-test-app/index.pl
mnagy: I'll create the application, no problem.

5.16/Dockerfile
mnagy: @jhadvig Can you please apply the following during your next update?

 - Remove the first TODO as I see you already applied it and also apply it in the second line that sets up the allow_build_interactivity,
 - Expand the second TODO as follows, so it is more verbose:
```
# TODO: Cleanup cpanp cache after cpanminus is installed?
```
 - remove the third one

5.16/test/run
jhadvig: @soltysh @mfojtik takeing back my +1
What if we decide to add more test-apps ? Like the have it like this(logic that can handle 1-n test apps) in other images I would rather stick to that. 
Agreed on that  we dont need to declare any array for that.

jhadvig:
@mfojtik @soltysh @bparees updated and rebased based on your comments. @mnagy will deliver the suitable replacement for the test-app.
PTAL

soltysh:
LGTM, except for that sample-app, once that will be delivered I'm OK with merging this.

mnagy:
@jhadvig I've updated the perl example application: https://github.com/mnagy/sample-perl-app

jhadvig:
@soltysh @mfojtik I've updated the PR with the test app that @mnagy provided. Thanks
PTAL :)

soltysh:
You still got my LGTM :smile:

mfojtik:
LGTM

mfojtik:
@jhadvig @mnagy need Jenkins and vagrant plugin update

mnagy:
@mfojtik @jhadvig done: https://github.com/openshift/vagrant-openshift/pull/239

PR #4, opened 2015-05-01
Support running Perl containers with a random uid

csrwng:
Makes it possible to run resulting images using any uid


csrwng:
@bparees - changes here are similar to the ones in the php image. ptal.

bparees:
ok to test

bparees:
add to whitelist

bparees:
lgtm, pending test results (if we can get the thing to test it)

bparees:
@csrwng humor me and close this PR and open it as a new one (make sure it gets a new PR#).  I think maybe i fixed the issue.

PR #5, opened 2015-05-04
Support running Perl containers with a random uid

csrwng:


csrwng:
@bparees reopened

bparees:
lgtm and looks like the test build is running. woo!

PR #6, opened 2015-05-05
merge env and run commands to reduce layers

bparees:

PR #7, opened 2015-05-06
Update image metadata to use LABEL

mfojtik:


mfojtik:
https://github.com/openshift/sti-ruby/issues/24

5.16/Dockerfile.rhel7
```diff
@@ -3,25 +3,29 @@ FROM openshift/base-rhel7
...
+      openshift.io/expose-services="8080:http" \
+      openshift.io/tags="builder,perl,perl516"
 
 RUN yum-config-manager --enable rhel-server-rhscl-7-rpms \
-	yum-config-manager --enable rhel-7-server-optional-rpms && \
-	yum install -y --setopt=tsflags=nodocs httpd24 perl516 perl516-mod_perl perl516-perl-CPANPLUS && \
-	yum clean all
+  yum-config-manager --enable rhel-7-server-optional-rpms && \
+  yum install -y --setopt=tsflags=nodocs httpd24 perl516 perl516-mod_perl perl516-perl-CPANPLUS && \
+  yum clean all
```
mnagy: Can you please use four spaces for better alignment?

mnagy:
This fails to build for the same reason as #6. Meaning the PATH is not set correctly and cpanm is not found because /opt/rh/perl516/root/usr/local/bin is not in PATH. Can you put ENV PATH on a separate line and test if that'll help?

Also, this will conflict with #6 so maybe coordinate with @bparees.

mfojtik:
@mnagy problem here is that we downgrade docker in jenkins and the old version of docker does not support having ENV vars on one line... This should be fixed when we get rid of downgrade. You can update the Jenkins job for perl image and remove the downgrade here (it does not make any sense there anyway as we are not pushing anything in that job).

mnagy:
LGTM

soltysh:
LGTM

soltysh:
Sorry, wrong button.

PR #17, opened 2015-06-02
Move installation of cpanm to assemble

mfojtik:


mfojtik:
@bparees @mnagy PTAL

mfojtik:
Locked to 1.7102 which is latest stable released version.

bparees:
lgtm.

mfojtik:
@hhorak FYI

Just to clarify this... This is not a problem-solver, just temporary workaround before we get RPM package we can use. Having this installation in assemble make us responsible for fixing security bugs along other things.

PR #18, opened 2015-06-05
Adding PerlPassEnv to pass DB specific variables to mod_perl

ewolinetz:


ewolinetz:
@bparees PTAL, variable names taken from their respective repository READMEs

bparees:
@ewolinetz it's a worthy start.  But of course if anyone uses a different database service name this won't be sufficient...  does our perl STI assembler allow you to supply your own conf file?

ewolinetz:
@bparees per our convo httpd.conf is now configured so that it can read in user provided .conf files.  It made sense to me to keep the two PerlPassEnv lines as a user wouldn't define the DATABASE_SERVICE_HOST and DATABASE_SERVICE_PORT in their DC so they wouldn't know to add them to their custom conf file...

bparees:
yup, keeping those as out of the box available make sense.

looks like it failed... i get a slightly different error when i try it, once that makes sense though:
httpd: Syntax error in -C/-c directive: Syntax error on line 16 of /opt/openshift/httpd.conf: No matches for the wildcard '*.conf' in '/opt/openshift/config', failing (use IncludeOptional if required)

ewolinetz:
Updated Include to be IncludeOptional and verified error scenario and positive scenario.  Also added another check in assemble due to avoid error where cfg folder is provided but is empty/doesn't include .conf files

bparees:
Still seems to be failing our tests...you can run "make test" to test this locally.

[INFO] Testing 'docker run' usage...


[INFO] Waiting for application container to start...


[INFO] Testing the HTTP connection (http://172.17.0.11:8080)...


[INFO] TEST FAILED (1)...

ewolinetz:
make test completed successfully locally for me.. 

[INFO] Testing 'sti usage'...
[INFO] Testing 'docker run' usage...
[INFO] Waiting for application container to start...
[Mon Jun 08 14:20:35.355285 2015] [so:warn] [pid 1] AH01574: module perl_module is already loaded, skipping
AH00558: httpd: Could not reliably determine the server's fully qualified domain name, using 172.17.0.5. Set the 'ServerName' directive globally to suppress this message
[INFO] Testing the HTTP connection (http://172.17.0.5:8080)...
[INFO] Cleaning up the test application...
20baa0f08452ba84ff4663497ce939b03ff2eddbe908772401933f9500348a3e
Error response from daemon: Cannot destroy container 20baa0f08452ba84ff4663497ce939b03ff2eddbe908772401933f9500348a3e: Failed to set container state to RemovalInProgress: Status is already RemovalInProgress
FATA[0000] Error: failed to remove one or more containers 
[INFO] Waiting for application container to start...
[Mon Jun 08 14:20:37.274376 2015] [so:warn] [pid 1] AH01574: module perl_module is already loaded, skipping
AH00558: httpd: Could not reliably determine the server's fully qualified domain name, using 172.17.0.6. Set the 'ServerName' directive globally to suppress this message
[INFO] Testing the HTTP connection (http://172.17.0.6:8080)...
[INFO] Cleaning up the test application...
419ef579e6fc37d411fe3ee7e02587c72e712e9547053329fa2d4f8954267984
Error response from daemon: Cannot destroy container 419ef579e6fc37d411fe3ee7e02587c72e712e9547053329fa2d4f8954267984: Failed to set container state to RemovalInProgress: Status is already RemovalInProgress
FATA[0000] Error: failed to remove one or more containers 
[INFO] All tests for the sample-test-app finished successfully....
[INFO] Cleaning up the test application image...
Untagged: openshift/perl-516-centos7-candidate-testapp:latest
Deleted: 140aadfb6bbe3233bc67f7558ec0768c9f2064324696b869396dbddbe131e4d9

[INFO] All tests finished successfully....

bparees:
rerunning the test build here:
https://ci.openshift.redhat.com/jenkins/job/sti-perl/41/

ewolinetz:
@bparees The fix for the rhel test failiing was included in the 5.16/contrib/httpd.conf file.  I'm not sure if this is the correct place for this or if it should be with the DocumentRoot definition

bparees:
@ewolinetz works for me, thanks for sorting it out!

PR #19, opened 2015-06-09
Remove source of .bashrc

mnagy:
The sti base image will take care of it through BASH_ENV.

This depends on https://github.com/openshift/sti-base/pull/33

### DIFF
--- a/5.16/.sti/bin/assemble
+++ b/5.16/.sti/bin/assemble
@@ -1,8 +1,5 @@
 #!/bin/bash
 
-# For SCL enablement
-source .bashrc
-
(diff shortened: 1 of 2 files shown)


mnagy:
@mfojtik

mfojtik:
have to be tested with perl-ex, nodejs-ex, rails-ex and php example.

mnagy:
@mfojtik where do those live? sti-perl has a test included

jhadvig:
@mnagy we dont have tests for the *-ex application, you will probably need to test the changed images manually.

PR #24, opened 2015-06-15
Updating to allow for dynamically generated env.conf file to expose ...

ewolinetz:
all container environment variables to mod_perl

### DIFF
--- a/5.16/.sti/bin/run
+++ b/5.16/.sti/bin/run
@@ -2,4 +2,8 @@
 
 set -e
 
+# Warning: 
+# Please note that this will pass all environment variables available within the container to mod_perl by means of PerlPassEnv in the env.conf file
+env | awk -F'=' '{print "PerlPassEnv "$1}' >> /opt/openshift/config/env.conf
+
 exec httpd -C 'Include /opt/openshift/etc/httpd.conf' -D FOREGROUND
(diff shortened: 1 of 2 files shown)


ewolinetz:
@bparees PTAL, post 3.0 candidate

bparees:
seems reasonable enough to me, did you do any research to see if there were any warnings about why not to do this?

ewolinetz:
probably not enough, from what I've gathered so far by default mod_perl automatically allows a few environment variables and anything past those need to be specified with PerlPassEnv or PassEnv.  The difference being at what point they are exposed to Perl, with PerlPassEnv exposing it as soon as possible, the former is exposed after the fixup phase (http://perl.apache.org/docs/1.0/guide/config.html#PerlSetEnv_and_PerlPassEnv)

5.16/.sti/bin/run
```diff
@@ -2,4 +2,6 @@
 
 set -e
 
+exec env | awk -F'=' '{print "PerlPassEnv "$1}' >> /opt/openshift/config/env.conf
```
mnagy: s/exec //

mnagy:
LGTM

mfojtik:
LGTM although security me is screaming now :-)

5.16/.sti/bin/run
```diff
@@ -2,4 +2,6 @@
 
 set -e
 
+env | awk -F'=' '{print "PerlPassEnv "$1}' >> /opt/openshift/config/env.conf
```
bparees: please throw a warning comment before this explaining that it's passing all env through to mod_perl.

bparees:
one nit and lgtm.

ewolinetz:
Addressed comment

PR #25, opened 2015-06-19
Randomize the UID for container that run the test

mfojtik:


mfojtik:
https://github.com/openshift/sti-base/issues/44

mnagy:
I think we should test it with the default user as well, not just a "randomized" uid.

mfojtik:
@mnagy what is the point of doing that? 'default' user is another 'random' user ;-)

mnagy:
@mfojtik Technically yes, but I think I'd be nice to cover the docker use-case of not using a --user flag to make sure that the default uids work.

mfojtik:
@mnagy i don't want to make our tests run longer just to prove that 'default' user will behave same as random user. The random user naturally has less privileges than default and we care only about random user here. If random user can't run the app, then we are broken. If that happens, we don't care if 'default' user has the same problem or not.

mnagy:
@mfojtik ok, I take that back. One more reason I had was that I thought we were doing some "default user" stuff in docker files, but I just realized that we do that in DB images.

LGTM

PR #31, opened 2015-06-29
Fix the label schema

mfojtik:


5.16/Dockerfile
```diff
@@ -33,9 +29,12 @@ RUN yum install -y \
     sed -i -f /opt/openshift/etc/httpdconf.sed /opt/rh/httpd24/root/etc/httpd/conf/httpd.conf  && \
     chmod -R og+rwx /opt/rh/httpd24/root/var/run/httpd && \
     mkdir /opt/openshift/config && \
-    chmod -R og+rwx /opt/openshift/config
```
bparees: /opt/openshit/config still needs to be chmod'd right?  it's being created one line up, as root.

mfojtik: yes, working on that now

mfojtik:
@bparees fixed, want to see if Jenkins is green

bparees:
you and me both.  the failure before was happening earlier so this won't have fixed that i don't think.

mfojtik:
oh you're right, the chown must be done after copy

5.16/Dockerfile
```diff
@@ -29,13 +25,16 @@ RUN yum install -y \
...
+    chmod -R og+rwx /opt/rh/httpd24/root/var/run/httpd
+
+# Each language image can have 'contrib' a directory with extra files needed to
+# run and build the applications.
+COPY ./contrib/ /opt/openshift
 
+# Drop the root user and make the content of /opt/openshift owned by user 1001
 USER 1001
+RUN chown -R 1001:1001 /opt/openshift
```
bparees: doesn't this need to come before USER 1001?

mfojtik: yeah

bparees:
sed: couldn't open file /opt/openshift/etc/httpdconf.sed: No such file or directory

bparees:
so that's another reason the COPY has to happen before the RUN :)

mfojtik:
@bparees I'm just on it :-) one sec

mfojtik:
(this time I will test it to not waste jenkins time ;-)

mfojtik:
@bparees done :-)

5.16/.sti/bin/run
```diff
@@ -4,6 +4,6 @@ set -e
 
 # Warning: 
 # Please note that this will pass all environment variables available within the container to mod_perl by means of PerlPassEnv in the env.conf file
-env | awk -F'=' '{print "PerlPassEnv "$1}' >> /opt/openshift/config/env.conf
+env | awk -F'=' '{print "PerlPassEnv "$1}' >> /opt/openshift/etc/httpd.d/env.conf
```
bparees: the point of this being in the config dir is it could be overwritten by the user supplying a env.conf file in their source repo in case they don't want all env variables being passed through.

(assemble copies contents into the config dir)

bparees: actually i guess that's not true since we overwrite it when run executes.

but that *should* have been the point :)

bparees: run should only write to env.conf if it doesn't exist.

that way users can supply their own.

mfojtik: @bparees fixed assemble.... we should have test case for that...

5.16/Dockerfile.rhel7
```diff
@@ -23,23 +23,23 @@ LABEL Component="perl516" \
...
-    chmod -R og+rwx /opt/rh/httpd24/root/var/run/httpd && \
-    mkdir /opt/openshift/config && \
-    chmod -R og+rwx /opt/openshift/config
+    scl enable perl516 -- cpanp 's conf allow_build_interactivity 0; s save system'
 
+# Each language image can have 'contrib' a directory with extra files needed to
+# run and build the applications.
+COPY ./contrib/ /opt/openshift
+
+# Drop the root user and make the content of /opt/openshift owned by user 1001
```
bparees: this comment needs to be moved down to the USER statement.  at least the first half of it.

5.16/.sti/bin/run
bparees: +1

5.16/Dockerfile.rhel7
mfojtik: I ment this comment to apply to the RUN as well, where you have to chown/chmod some dir in order to drop the user privileges.

mfojtik: I will reword it

bparees:
lgtm.

PR #39, opened 2015-07-24
Perl 5.20

ppisar:
This adds support for Perl 5.20 as provided by rh-perl520 SCL. It's clone of 5.16 directory with changes needed for Perl 5.20, with collection enablement using bash environment variable, and with installing other packages needed to support XS Perl modules.


bparees:
[test]

PR #47, opened 2015-09-05
make source owned by gid 0

bparees:


jhadvig:
@bparees why do we need to change 0 group ? Do we need to change it in other httpd based images as well ?

bparees:
I'm working on it :)

The uid for the container is always a member of group 0, so this will
ensure you can exec into the container and edit source for hot
reload/debug.

All the s2i images will be updated when I'm done.

Ben Parees | OpenShift

jhadvig:
Sweeet :+1: 

Thought you guys in US have weekend as well :smile:

bparees:
the weekend is when i actually have free time to do this stuff :)


On Sat, Sep 5, 2015 at 8:04 PM, Jakub Hadvig <notifications@github.com>
wrote:

> Sweeet [image: :+1:]
>
> Thought you guys in US have weekend as well [image: :smile:]
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openshift/sti-perl/pull/47#issuecomment-138010569>.
>



-- 
Ben Parees | OpenShift

jhadvig:
good rule :satisfied:... will write that down

5.16/s2i/bin/assemble
```diff
@@ -4,6 +4,8 @@ set -e
 
 echo "---> Installing application source"
 cp -Rf /tmp/src/. ./
+chgrp -R 0 ./
```
jhadvig: Should this be done in Dockerfile, by setting the setGID on the folder ?
I could imagine that something like `chmod g+s ${HOME}` could go to the sti-base PR you opened, at the end of the neverending RUN instruction :smile:, cause I see that the testi is failing now in this command due to the permissions. 
Handing this logic to the sti-base would be more suitable, since all the lang. image PR have this two lines in common, except wildfly.
On the other hand, wouldn't the wildfly exception prevent from handling this issue in the Dockerfile ?

bparees: The sticky bit might work, I'll try it.

Ben Parees | OpenShift

bparees: https://serverfault.com/questions/10353/what-is-the-sticky-bit-in-unix-file-systems-when-is-it-used

i don't think the sticky bit solves this problem because it doesn't make the source files themselves editable or deletable, it just allows random users to create new files.

hhorak:
Please, provide proper reasoning why this change is done.

hhorak:
@bparees ^

jhadvig:
@hhorak https://github.com/openshift/sti-perl/pull/47#issuecomment-138010007

5.16/s2i/bin/assemble
jhadvig: @bparees your right, didn't noticed that limitation. Pity, I would like to see those kind of changes in Dockerfiles :confused:

hhorak:
@bparees "The uid for the container is always a member of group 0" -- is this just a convention or is it ensured somewhere? in OpenShift? generally in docker?

bparees:
@hhorak i believe the behavior comes from docker:
$ docker run -itu 1001 centos groups
root

PR #49, opened 2015-09-23
Add support for Hot Deploy

mnagy:
Tested.

### DIFF
--- a/5.16/s2i/bin/run
+++ b/5.16/s2i/bin/run
@@ -8,4 +8,13 @@ if [ ! -f /opt/app-root/etc/httpd.d/env.conf ]; then
   env | awk -F'=' '{print "PerlPassEnv "$1}' > /opt/app-root/etc/httpd.d/env.conf
 fi
 
+# Enable automatic reloading. This can be useful for debugging an application.
+PERL_APACHE2_RELOAD=${PERL_APACHE2_RELOAD:-}
+if [[ "${PERL_APACHE2_RELOAD,,}" == "true" ]]; then
+  cat > /opt/app-root/etc/httpd.d/50-autoreload.conf <<EOF
+PerlModule Apache2::Reload
+PerlInitHandler Apache2::Reload
+EOF
+fi
+
 exec httpd -C 'Include /opt/app-root/etc/httpd.conf' -D FOREGROUND
--- a/5.20/s2i/bin/run
+++ b/5.20/s2i/bin/run
@@ -8,4 +8,13 @@ if [ ! -f /opt/app-root/etc/httpd.d/env.conf ]; then
   env | awk -F'=' '{print "PerlPassEnv "$1}' > /opt/app-root/etc/httpd.d/env.conf
 fi
 
+# Enable automatic reloading. This can be useful for debugging an application.
+PERL_APACHE2_RELOAD=${PERL_APACHE2_RELOAD:-}
+if [[ "${PERL_APACHE2_RELOAD,,}" == "true" ]]; then
+  cat > /opt/app-root/etc/httpd.d/50-autoreload.conf <<EOF
+PerlModule Apache2::Reload
+PerlInitHandler Apache2::Reload
+EOF
+fi
+
 exec httpd -C 'Include /opt/app-root/etc/httpd.conf' -D FOREGROUND
--- a/README.md
+++ b/README.md
@@ -189,3 +189,7 @@ file inside your source code repository.
 
     This variable specifies a mirror URL which will used by cpanminus to install dependencies.
     By default the URL is not specified.
+
+* **PERL_APACHE2_RELOAD**
+
+    Set this to "true" to enable automatic reloading of modified Perl modules.


README.md
```diff
@@ -189,3 +189,7 @@ file inside your source code repository.
 
     This variable specifies a mirror URL which will used by cpanminus to install dependencies.
     By default the URL is not specified.
+
+* **PERL_APACHE2_RELOAD**
+
+    Turn on automatic reloading of modified Perl modules. The default value is `false`.
```
jhadvig: This variable is not set in this image, just in the dencer-ex so doesn't make sense to write about the default value.
Rather write how can you enable the hot deploy (eg. `oc new-app perl-image -e PERL_APACHE2_RELOAD=true`) and what is the expected values of that variable

jhadvig: nm the default value, keep that

rhcarvalho: default=\<not set\> != default=false

Same for the other existing env var. IMHO the doc is lacking in clarity.

mnagy:
PTAL

5.16/s2i/bin/run
```diff
@@ -8,4 +8,12 @@ if [ ! -f /opt/app-root/etc/httpd.d/env.conf ]; then
   env | awk -F'=' '{print "PerlPassEnv "$1}' > /opt/app-root/etc/httpd.d/env.conf
 fi
 
+# Enable automatic reloading. This can be useful for debugging an application.
+if [[ -n "${PERL_APACHE2_RELOAD:-}" && "${PERL_APACHE2_RELOAD,,}" != "false" && "${PERL_APACHE2_RELOAD}" != "0" ]]; then
+  cat > /opt/app-root/etc/httpd.d/autoreload.conf <<EOF
```
mfojtik: prefix the filename with number, I think @ppisar suggested that and I agree...

5.20/s2i/bin/run
```diff
@@ -8,4 +8,12 @@ if [ ! -f /opt/app-root/etc/httpd.d/env.conf ]; then
   env | awk -F'=' '{print "PerlPassEnv "$1}' > /opt/app-root/etc/httpd.d/env.conf
 fi
 
+# Enable automatic reloading. This can be useful for debugging an application.
+if [[ -n "${PERL_APACHE2_RELOAD:-}" && "${PERL_APACHE2_RELOAD,,}" != "false" && "${PERL_APACHE2_RELOAD}" != "0" ]]; then
+  cat > /opt/app-root/etc/httpd.d/autoreload.conf <<EOF
```
mfojtik: same here

5.20/s2i/bin/run
```diff
@@ -8,4 +8,12 @@ if [ ! -f /opt/app-root/etc/httpd.d/env.conf ]; then
   env | awk -F'=' '{print "PerlPassEnv "$1}' > /opt/app-root/etc/httpd.d/env.conf
 fi
 
+# Enable automatic reloading. This can be useful for debugging an application.
+if [[ -n "${PERL_APACHE2_RELOAD:-}" && "${PERL_APACHE2_RELOAD,,}" != "false" && "${PERL_APACHE2_RELOAD}" != "0" ]]; then
```
mfojtik: omg :-) can we just check if the variable is set? simple `[ -v PERL_APACHE2_RELOAD ]` will do.

mfojtik: or keep it as it is but you will have to explain all options and possible values in documentation, which in my opinion doesn't make any sense.

README.md
```diff
@@ -189,3 +189,7 @@ file inside your source code repository.
 
     This variable specifies a mirror URL which will used by cpanminus to install dependencies.
     By default the URL is not specified.
+
+* **PERL_APACHE2_RELOAD**
+
+    Set this to anything other than "false" or "0" to enable automatic reloading of modified Perl modules.
```
rhcarvalho: empty, false , 0?

5.20/s2i/bin/run
mfojtik: the only added value I see here (to check the variable value) is to have template parameter that will set this to true/false (which will most likely never happen). In that case, we should just have PERL_APACHE2_RELOAD=true as the only valid way to "enable" this. Anything else as a value or the non-existence of this variable means you don't want this.... having this opposite logic here is hard to read, maintain and complex to document.

mfojtik: @mnagy  @rhcarvalho please stop bike-shedding around this and decide one value that will mean you want enable this thing and move on.

jhadvig: `true`

README.md
mfojtik: This should be:

```
Set this to `true` if you want to enable automatic code reloading for your Perl code
```

5.20/s2i/bin/run
mnagy: @mfojtik this was agreed upon yesterday with @bparees and @rhcarvalho. I do not care about any way it will work. However, @bparees had a good point against `[ -v PERL_APACHE2_RELOAD ]`, that once defined, you can't disable it by setting it to another value (e.g. `false`). What Ben also didn't like was that false would turn it on.

mfojtik: @bparees I think the simplest possible thing here is to document that PERL_APACHE2_RELOAD=true turn this on, everything else disables this...

bparees: simplest but not best.  we had a lengthy discussion on this yesterday.  also you're contradicting your own complaint about bikeshedding :)

what i requested was that having the variable be not-set, set to false(case insensitive), or "0" would not enable reload.

which is what we have here, so the bikeshed looks awesome to me :)  :+1: @mnagy

README.md
bparees: "empty" is not the same as "unset".  

https://gist.github.com/bparees/1e7bbde35025bda63c79
(though i'm not sure @mnagy's tests currently differentiate)

@mfojtik and then someone sets the value to "1" and wonders why it doesn't work.

mfojtik: @bparees if we say in documentation that you have to set this to `true` in order to enable this and you set it to 1 it means you don't read documentation. We can protect users from doing stupid things. Also what if I set this to "disable" ? Should we also add this to list of negations? Allow just "true" and then you have super-simple `if`, no negations in documentatiom etc.

bparees: @mfojtik that is fine with me.

mnagy:
Updated. I added the empty option into README.md to be more accurate and added numbering for the httpd conf.

mnagy:
Updated again, PTAL.

jhadvig:
LGTM

bparees:
lgtm

PR #51, opened 2015-09-29
use env variable for script copy location

bparees:


jhadvig:
Since the `STI_SCRIPTS_PATH` is set to `image:///usr/local/sti`, which is on a proper PATH this won't work.

bparees:
@jhadvig see https://github.com/openshift/sti-base/pull/60/files#diff-b0556bed298b3a4fb269eb10d69c91a5R14

bparees:
STI_SCRIPTS_PATH!=STI_SCRIPTS_URL

jhadvig:
LGTM

PR #53, opened 2015-10-09
Configure Apache to write access logs to stdout instead of file.

php-coder:
Copy the behavior of php images: https://github.com/openshift/sti-php/pull/64

PTAL @mfojtik @rhcarvalho

BTW, should we add the same Logs section as we have for PHP? Right now we don't have such section but IMHO it may be useful.


rhcarvalho:
Thanks @php-coder LGTM. Please add the Logs documentation as well, just like we have for PHP.

php-coder:
> Please add the Logs documentation as well, just like we have for PHP.

Done.

mnagy:
LGTM

PR #60, opened 2015-11-06
Fix executing script installed from CPAN

ppisar:
It's not possible to execute scripts coming from CPAN. This is because the ~/extlib/bin is not in the PATH. Also ~/extlib/lib/perl5 must be added to Perl's include through environment variable in order to such script executed via kernel's execve() know where CPAN Perl modules  are installed.

This also augments the test/run script so that new tests can be added later.

### DIFF
--- a/5.20/s2i/bin/run
+++ b/5.20/s2i/bin/run
@@ -2,6 +2,12 @@
 
 set -e
 
+# CPAN can install scripts. They should be available from mod_perl too.
+export PATH=${PATH}:/opt/app-root/src/extlib/bin
+# And we have to set Perl include path too because mod_perl's PerlSwitches
+# does not apply to them.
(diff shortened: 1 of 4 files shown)


5.20/s2i/bin/run
```diff
@@ -2,6 +2,12 @@
 
 set -e
 
+# CPAN can install scripts. They should be available from mod_perl too.
+export PATH=${PATH}:${HOME}/extlib/bin
```
mfojtik: do not use $HOME variable, this might change unpredictable way, we prefer absolute path

ppisar: On Tue, Nov 10, 2015 at 05:02:48AM -0800, Michal Fojtik wrote:
> do not use $HOME variable, this might change unpredictable way, we prefer
> absolute path
> 
I corrected the patch not to use the HOME variable and rebased the branch.

PR #62, opened 2015-12-01
Change RH required labels based on updated spec

hhorak:
For full spec, see https://github.com/projectatomic/ContainerApplicationGenericLabels/blob/master/vendor/redhat/labels.md

### DIFF
--- a/5.16/Dockerfile.rhel7
+++ b/5.16/Dockerfile.rhel7
@@ -7,17 +7,18 @@ EXPOSE 8080
 ENV PERL_VERSION=5.16 \
     PATH=$PATH:/opt/rh/perl516/root/usr/local/bin
 
-LABEL io.k8s.description="Platform for building and running Perl 5.1.6 applications" \
+LABEL summary="Platform for building and running Perl 5.1.6 applications" \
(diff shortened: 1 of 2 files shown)


ppisar:
On Tue, Dec 01, 2015 at 02:34:42AM -0800, Honza Horak wrote:
>   * Change RH required labels based on updated https://github.com/projectatomic/ContainerApplicationGenericLabels/blob/master/vendor/redhat/labels.md
> 
I'm fine with this change. What's OpenShift people's opinion? Is renaming the `Name' label to `name' acceptable for you?

-- Petr

mfojtik:
@ppisar the only labels we care about are the k8s and openshift namespaced labels, so we have no problem changing that.

@hhorak is this still up to date with latest atomic 'standards'? If so I can merge it. Also @rhcarvalho was doing some cleanup of layers by combining the LABEL commands together, maybe worth sync with them  about that (and the other PR).

rhcarvalho:
@mfojtik all equivalent PRs with the same content in other image repos were either closed or merged and reverted, because the build system doesn't support the new labels yet.

hhorak:
Closing, since osbs does not allow these new specs yet.

PR #64, opened 2015-12-16
Add support for REGISTRY parameter

mfojtik:


ppisar:
On Wed, Dec 16, 2015 at 04:03:20AM -0800, Michal Fojtik wrote:
>   https://github.com/openshift/sti-perl/pull/64
> 
> -- Commit Summary --
> 
>   * Add support for REGISTRY parameter
> 
No problem with it. I only wonder why you used

+    if [[ ! -z "${REGISTRY}" ]]; then

instead of

+    if [[ -n "${REGISTRY}" ]]; then

-- Petr

mfojtik:
@ppisar yeah, that is better, I will change it. Thanks

PR #65, opened 2015-12-18
link to openshift docs

bparees:


ppisar:
On Fri, Dec 18, 2015 at 12:35:35PM -0800, Ben Parees wrote:
>   https://github.com/openshift/sti-perl/pull/65
> 
> -- Commit Summary --
> 
>   * link to openshift docs
> 
> -- File Changes --
> 
>     M 5.16/README.md (2)
> 
What about 5.20/README.md? Is the text applicable to 5.16 only?

-- Petr

bparees:
@ppisar actually i updated the wrong readme.  i've been trying to just update the top-level readme in all the repos but i touched the wrong one in this case.  Thanks for catching it.

fixing here:
https://github.com/openshift/sti-perl/pull/66

PR #67, opened 2016-01-12
Verify all installed packages using rpm -V

mfojtik:


mfojtik:
[test]

mfojtik:
@bparees 

package rh-perl520-devel is not installed
ERROR: perl-rhel7 failed testing.

First bug catched...

mfojtik:
@ppisar do you know why that package does not exists?

ppisar:
On Tue, Jan 12, 2016 at 03:50:38AM -0800, Michal Fojtik wrote:
> @ppisar do you know why that package does not exists?
> 
I know. That's because it's a typo. The "rh-perl520-devel" package should be
spelled "rh-perl520-perl-devel".

The reason why it the yum command succeeds even with the typo is because yum
reports success if at least one package is successfully installed.

The fix is to s/rh-perl520-devel/rh-perl520-perl-devel/ in 5.20/Dockerfile*.

-- Petr

ppisar:
On Tue, Jan 12, 2016 at 03:02:11AM -0800, Michal Fojtik wrote:
> 
>   * Verify all installed packages using rpm -V
> 
I don't understand the reason for this change. It obiously verifies only the
explicitely installed packages. It does not cover their implicit dependencies.
What's the purpose of it? Morever the yum argument can be an RPM Provide or
a file or a package group. These specification are not acceptable by rpm
command.

-- Petr

mfojtik:
@ppisar the reason why we do this is that we want to make sure that all rpms we explicitly mark to install are really installed and are not missing in the final image. This already happened for s2i-python image where the nss_wrapper was not installed even in the RHEL variant of the image.

The good solution for this is to have an option for `yum` to fail when one of the packages are not installed. But this option is not there yet and we don't want to go inspect the jenkins logs or OSBS logs if all packages are properly installed.

ppisar:
On Tue, Jan 12, 2016 at 05:55:59AM -0800, Michal Fojtik wrote:
> @ppisar the reason why we do this is that we want to make sure that all rpms
> we explicitly mark to install are really installed and are not missing in
> the final image. This already happened for s2i-python image where the
> nss_wrapper was not installed even in the RHEL variant of the image.
> 
I see. Have you considered plain "rpm -q". It's remarkably faster than "rpm
-V". Or "rpm -q --whatprovides" that would accept RPM Provides and files too.
The only uncovered yum specifier would be package group.

ppisar:
On Tue, Jan 12, 2016 at 02:43:39PM +0100, Petr Pisar wrote:
> On Tue, Jan 12, 2016 at 03:50:38AM -0800, Michal Fojtik wrote:
> > @ppisar do you know why that package does not exists?
> > 
> I know. That's because it's a typo. The "rh-perl520-devel" package should be
> spelled "rh-perl520-perl-devel".
> 
> The reason why it the yum command succeeds even with the typo is because yum
> reports success if at least one package is successfully installed.
> 
> The fix is to s/rh-perl520-devel/rh-perl520-perl-devel/ in 5.20/Dockerfile*.
> 
Should I write a fix as a separate pull request that you apply before the
verification commit, or will you handle it in your current pull request?

-- Petr

mfojtik:
@ppisar I can fix that in this PR, thanks.

PR #69, opened 2016-02-12
install gd-devel package

bparees:
fixes https://github.com/openshift/sti-perl/issues/68

### DIFF
--- a/5.16/Dockerfile
+++ b/5.16/Dockerfile
@@ -16,7 +16,7 @@ LABEL io.k8s.description="Platform for building and running Perl 5.1.6 applicati
 
 # TODO: Cleanup cpanp cache after cpanminus is installed?
 RUN yum install -y centos-release-scl && \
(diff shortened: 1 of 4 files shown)


bparees:
@nhr ptal

ppisar:
I'm against this commit. There is no reason why to install exactly this package. In general I'm against adding arbitrary package.

ppisar:
If you want put this OpenShift work around somewhere, put into base image. The gd requirement is nothing specific to Perl. You can have the same request in PHP, Python etc.

bparees:
@ppisar some packages are more commonly used by some frameworks than others.  eg imagemagick is a common php dep, even though it's not a php package, so i would install that in the php image, not the base image, at least as a starting point.

Whether this package falls in that category or not i'm less sure since i'm not a perl dev.

I'm ok w/ putting it in the base image instead, but first i'd like a little more confirmation that it is indeed commonly used by php/ruby/python/nodejs.

bparees:
this has been requested for python also so we'll add it to sti-base.

Issue #72, opened 2016-02-25
httpd log file /tmp_error log should be directed to STDERR


nhr:
Presently we direct [some httpd logging to STDOUT](https://github.com/openshift/sti-perl/blob/master/5.20/contrib/etc/httpdconf.sed#L7), however we do not expose error messages in a similar fashion. This complicates the process of troubleshooting Perl applications, as most of the runtime errors are being sent to the error_log file.

mfojtik:
both logs should log into stdout, this is a bug.

rhcarvalho:
Looks like this applies to PHP as well...

- https://github.com/openshift/sti-perl/pull/53
- https://github.com/openshift/sti-php/pull/64

Shall we send both to `stdout`, or distinguish `stdout` / `stderr`?

```
s%^ErrorLog "logs/error_log"%ErrorLog "|/usr/bin/cat"%
s%CustomLog "logs/access_log"%CustomLog "|/usr/bin/cat"%
```

@php-coder including you here since you worked on this before.
@mfojtik @bparees @nhr @php-coder WDYT? I would prefer to distinguish the streams.

php-coder:
I'm voting for distinguishing them too. Interesting to see how will you do this.

mfojtik:
We can prefix them, I assume httpd allows to set custom formatter.

bparees:
prefixing works for me.

php-coder:
Ideally, is to have a real stdout/stderr differentiation, it allows to programmatically distinguish them (like this: https://github.com/docker/docker/issues/7440#issuecomment-51365165)

php-coder:
I didn't try, but maybe the following would work:
```
s%^ErrorLog "logs/error_log"%ErrorLog "|perl -wlne 'print STDERR'"%
```

ppisar:
On Mon, Feb 29, 2016 at 02:54:58AM -0800, Rodolfo Carvalho wrote:
> 
> Shall we send both to `stdout`, or distinguish `stdout` / `stderr`?
> 
No. Access log to stdout, error log to stderr. See run-httpd24.sh from
httpd24:

~~~~
sed -ri ' s!^(\s*CustomLog)\s+\S+!\1 /proc/self/fd/1!g; s!^(\s*ErrorLog)\s+\S+!\1 /proc/self/fd/2!g;' /opt/rh/httpd24/root/etc/httpd/conf/httpd.conf
~~~~

php-coder:
[...]

PR #73, opened 2016-03-02
Log httpd log_error to stderr

ppisar:
This patch set fixes issue #72.

### DIFF
--- a/5.20/test/run
+++ b/5.20/test/run
@@ -160,6 +160,26 @@ test_response() {
   return $result
 }
 
+# Match PCRE regular expression against container standard output.
+# First argument is the PCRE regular expression.
+# It expects standard output in ${tmp_dir}/out file.
+test_stdout() {
+  local regexp="$1"
+  local output="${tmp_dir}/out"
+  info "Testing the container standard output for /${regexp}/"
+  grep -qP -e "$regexp" "$output";
+}
+
+# Match PCRE regular expression against container standard error output.
+# First argument is the PCRE regular expression.
+# It expects error output in ${tmp_dir}/err file.
+test_stderr() {
+  local regexp="$1"
+  local output="${tmp_dir}/err"
+  info "Testing the container error output for /${regexp}/"
+  grep -qP -e "$regexp" "$output";
+}
+
 test_connection() {
     test_response '/' 200 ''
 }
@@ -185,6 +205,8 @@ test_application() {
 # must return (or terminate with) non-zero value to report an failure,
 # 0 otherwise.
 # Third argument is s2i --env argument value.
+# The test function have available container ID in $cid_file, container output
+# in ${tmp_dir}/out, countainer stderr in ${tmp_dir}/err.
 do_test() {
     test_name="$1"
     test_function="$2"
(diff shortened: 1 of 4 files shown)


mfojtik:
LGTM

rhcarvalho:
@ppisar nice patch :clap: ! Thanks! :)


@mfojtik the original issue was left open. BTW we need this for PHP and perhaps other httpd-based images as well...

5.20/test/run
```diff
@@ -185,6 +205,8 @@ test_application() {
 # must return (or terminate with) non-zero value to report an failure,
 # 0 otherwise.
 # Third argument is s2i --env argument value.
+# The test function have available container ID in $cid_file, container output
+# in ${tmp_dir}/out, countainer stderr in ${tmp_dir}/err.
```
rhcarvalho: Minor nit: s/countainer/container

rhcarvalho:
Well, too late, but worth asking: could we get the tests for both image versions? (5.16 is missing)

ppisar:
On Wed, Mar 02, 2016 at 02:51:41PM -0800, Rodolfo Carvalho wrote:
> Well, too late, but worth asking: could we get the tests for both image
> versions? (5.16 is missing)
>
The 5.20 has different test "frame work". 5.16 tests are not extendable.

Actually the tests should be rewritten to something more suitable for testing
that supports nesting subtests, reporting failure reason, parallel testing
etc. E.g. Perl's Test::More. But I don't know if that were acceptable for
OpenShift. I've never seen what OpenShift test environment provides,
guarantees or expects.

-- Petr

rhcarvalho:
> Actually the tests should be rewritten to something more suitable for testing
> that supports nesting subtests, reporting failure reason, parallel testing
> etc.

I totally agree :+1:

The tests in this repository are OpenShift-agnostic, they require just `docker`.

However, we need to take a more holistic view when talking about a change like rewriting.
Historically, they started as a "common" "framework" that is generated by `s2i create`. That command creates a scaffolding for creating new S2I builder images.

Through the past year, when working on the builder images and trying to improve the tests, I was reminded to keep the consistency among the images we provide, which is good in one hand, but hard to maintain over time. And as you just pointed out, Perl tests for 5.16 and 5.20 went out of sync...

(and there is really another problem in there, copy-paste...)

IMHO Bash is great for short scripts, but is a very poor tool for writing the day-after-day more complex scripts that empower the images, and very poor for writing tests.

Even though Perl's Test::More is very cool, I doubt we have enough Perl know-how across the team to make it the "new language" for the scripts in all images.

Given what's in the base OS image, I'd say our choices are Bash, Perl, Python.

mfojtik:
@rhcarvalho the tests also depend on `s2i` binary.

mfojtik:
@rhcarvalho also as you said, few of us have confidence in perl or python or ruby to start writing tests for images. We do all do Go, but that leaves SCL folks out of loop because that is something they don't do. So even it sound silly, the "bash" is the only common framework we share currently :-) And I'm OK improving it or making the testing framework better.

PR #76, opened 2016-04-06
revert logging stderr to stdout.

bparees:


bparees:
@mfojtik @ppisar @hhorak ptal, this reverts the stderr logging behavior for now because we need docker1.8 compatibility for OSE.

bparees:
(intentionally left as separate commits for easer re-reversion in the future, we will want to retain the typo fixes)

mnagy:
Why do you revert typo fixes and applying them back? Can't you just revert two commits?

bparees:
no because the typo fix commit causes conflicts that prevent reverting the other commits.  or my git-fu is weak.

mnagy:
I've only seen one conflict with the 5.20/test/run command. You can just resolve it manually, add it and commit. Then you'll end up with just two revert commits.

bparees:
@mnagy if you have a better approach, please just create the pull.

mnagy:
Done: https://github.com/openshift/sti-perl/pull/77

ppisar:
Please consider my pull request #78 instead fo mfojtik's #77.

PR #77, opened 2016-04-06
Revert logging to stderr changes that break us for Docker 1.8

mnagy:
This is an attempt that should be a bit lighter than #76. We can just revert these two commits when we want to bring back the changes.

@mfojtik @ppisar @hhorak @bparees


bparees:
lgtm, thanks @mnagy

mnagy:
Closing in favour of #78

Issue #75, opened 2016-03-24
httpd: could not open error log file /proc/self/fd/2


hhorak:
When I build the current version of the source here and run the current version of the test, I get this error:
(13)Permission denied: AH00091: httpd: could not open error log file /proc/self/fd/2.

It works fine when trying to go back before this commit:
https://github.com/openshift/sti-perl/commit/3c704ab239b714488e25542e707b9f79ba7f24c0

So, it seems to me like PR #73 is not working in all cases. I don't understand why it works for someone else though.

ppisar:
I cannot do anything with this. I don't manage the Jenkins serversto explore what's the difference.

ppisar:
It looks like it fails with docker-1.8.2-10.el7.x86_64 that is the latest stable version in RHEL-7. It works with 1.9.1 that can be found in Fedora.

OpenShift, what docker version do you use?

hhorak wants to revert the feature of logging to stderr. Is that acceptable solution for OpenShift?

hhorak:
As @ppisar says, the perl image as it is now does not work on up2date RHEL, not even on up2date CentOS.

mfojtik:
@bparees setting P1 as the image seems to be broken on RHEL7

@hhorak @ppisar are you guys working on fix?

mfojtik:
@ppisar i'm fine with reverting logging to stderr if that breaks RHEL7

bparees:
@mfojtik @ppisar @hhorak agree i'm fine w/ reverting it.  could we try another solution like using a pipe (https://httpd.apache.org/docs/1.3/logs.html#piped)  and sending the log through cat or tail or something?

ppisar:
On Mon, Apr 04, 2016 at 02:05:45PM -0700, Ben Parees wrote:
>    @mfojtik @ppisar @hhorak agree i'm fine w/ reverting it. could we try
>    another solution like using a pipe
>    (https://httpd.apache.org/docs/1.3/logs.html#piped) and sending the log
>    through cat or tail or something?
> 
What's the additional value of sending logs into another process? At the end
you need to write errors into stderr. It does not matter whether it's httpd or
another process that gets permission denial.

Could you explain why current code works in your Jenkins instance? Is that
because you use newer docker that is not yet available in RHEL/CentoOS?

I looked into httpd Dockerfile and it actually does not send errors to stderr
even though the commit message says so. I need to ask it's maintainer what
it means.

ppisar:
[...]

PR #78, opened 2016-04-07
Log errors to stdout

ppisar:
This addresses issue #75 by log errors to standard output instead of storing them into a file. That was the main concern of the original issue #72.

It also preserves tests to make sure the errors are logged to standard output and to provide the useful test framework features.

Once support for docker 1.8 is dropped, reverting this tiny change will provide logging to error output.


mnagy:
LGTM

bparees:
thanks @ppisar!

bparees:
@ppisar this still fails on docker 1.8.

bparees:
```
(13)Permission denied: AH00091: httpd: could not open error log file /proc/self/fd/1.
AH00015: Unable to open logs
```

ppisar:
Please try <https://github.com/openshift/sti-perl/pull/80>. It should send access log a error log to stdout and it should work even with 1.8.2 docker.

PR #79, opened 2016-04-11
revert stderr behavior so it works on docker1.8

bparees:


bparees:
@ppisar @tdawson fyi.

PR #80, opened 2016-04-12
Log to stdout via cat

ppisar:
Docker 1.8.2 cannot open /proc/self/fd files. The only way how to
write to standard output is to write to already opened descriptor.

In order to obtain httpd errors from docker client, we send error to
standard output too.

I've checked it with 1.8.2 and 1.9.1 this time.

<https://github.com/openshift/sti-perl/pull/78>


bparees:
lgtm

PR #81, opened 2016-04-21
Add server limit logic

mnagy:
For trello card https://trello.com/c/HbqjUI5y/881-3-update-perl-image-to-autoconfigure-based-on-available-memory

### DIFF
--- a/5.16/s2i/bin/run
+++ b/5.16/s2i/bin/run
@@ -17,4 +17,24 @@ PerlInitHandler Apache2::Reload
 EOF
 fi
 
+export_vars=$(cgroup-limits) ; export $export_vars
+
+# Validate server limit option
+if [[ ! "${HTTPD_MAX_REQUEST_WORKERS:-1}" =~ ^[1-9][0-9]*$ || "${HTTPD_MAX_REQUEST_WORKERS:-1}" -gt 20000 ]]; then
+  echo "HTTPD_MAX_REQUEST_WORKERS needs be an integer between 1 and 20000"
+  exit 1
+fi
+
+# If there is no server limit specified, try to guess the best value
+if [ -z "${HTTPD_MAX_REQUEST_WORKERS:-}" ]; then
+  MAX_SERVER_LIMIT=$(((MEMORY_LIMIT_IN_BYTES/1024/1024 - 30) / 7))
+  MAX_SERVER_LIMIT=$((MAX_SERVER_LIMIT > 0 ? MAX_SERVER_LIMIT : 1))
+  export HTTPD_MAX_REQUEST_WORKERS=$((MAX_SERVER_LIMIT > 256 ? 256 : MAX_SERVER_LIMIT))
+fi
+
+export HTTPD_START_SERVERS=${HTTPD_START_SERVERS:-8}
+export HTTPD_MAX_SPARE_SERVERS=$((HTTPD_START_SERVERS+10))
+
+envsubst < /opt/app-root/etc/httpd.d/50-mpm.conf.template > /opt/app-root/etc/httpd.d/50-mpm.conf
+
 exec httpd -C 'Include /opt/app-root/etc/httpd.conf' -D FOREGROUND
--- a/5.16/README.md
+++ b/5.16/README.md
@@ -102,3 +102,13 @@ file inside your source code repository.
 * **PERL_APACHE2_RELOAD**
 
     Set this to "true" to enable automatic reloading of modified Perl modules.
+
+* **HTTPD_START_SERVERS**
+
+    The [StartServers](https://httpd.apache.org/docs/2.4/mod/mpm_common.html#startservers)
+    directive sets the number of child server processes created on startup. Default is 8.
+
+* **HTTPD_MAX_REQUEST_WORKERS**
+
+    Number of simultaneous requests that will be handled by Apache. The default
+    is 256, but it will be automatically lowered if memory is limited.
--- a/5.16/contrib/etc/httpd.d/50-mpm.conf.template
+++ b/5.16/contrib/etc/httpd.d/50-mpm.conf.template
@@ -1,0 +1,13 @@
+<IfModule mpm_prefork_module>
+    # This value should mirror what is set in MinSpareServers.
+    StartServers            ${HTTPD_START_SERVERS}
+    MinSpareServers         ${HTTPD_START_SERVERS}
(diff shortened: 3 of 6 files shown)


ppisar:
What's the purpose of the export_vars variable?

mnagy:
@ppisar it's there just so we don't export in one command. That way, if the cgroup script fails, the whole script will fail.

mnagy:
@bparees @mfojtik PTAL.

This still needs docs.

mnagy:
Added stuff to README.md

5.16/s2i/bin/run
```diff
@@ -17,4 +17,30 @@ PerlInitHandler Apache2::Reload
...
+if [ "${NO_MEMORY_LIMIT:-}" != "true" ]; then
+  # We need 30 MiB as a base and 7 MiBs per server. Bash works with whole
+  # numbers here, but for our purposes, it is sufficient.
+  MAX_SERVER_LIMIT=$(((MEMORY_LIMIT_IN_BYTES/1024/1024 - 30) / 7))
+fi
+
+# Cap the server limit to the memory, but only if it was not specified manually
+if [ "${HTTPD_SERVER_LIMIT:-20000}" -gt "${MAX_SERVER_LIMIT}" ]; then
+  HTTPD_SERVER_LIMIT=${MAX_SERVER_LIMIT}
+fi
```
bparees: if the user sets HTTPD_SERVER_LIMIT explicitly, you should respect the value, not cap it at MAX_SERVER_LIMIT.

mnagy:
@bparees I reworked the whole logic, making the whole thing simpler, PTAL

5.16/s2i/bin/run
```diff
@@ -17,4 +17,23 @@ PerlInitHandler Apache2::Reload
...
+# Validate server limit option
+if [[ ! "${HTTPD_SERVER_LIMIT:-1}" =~ ^[1-9][0-9]*$ || "${HTTPD_SERVER_LIMIT:-1}" -gt 20000 ]]; then
+  echo "HTTPD_SERVER_LIMIT needs be an integer between 1 and 20000"
+  exit 1
+fi
+
+# If there is no server limit specified, try to guess the best value
+if [ -z "${HTTPD_SERVER_LIMIT:-}" ]; then
+  MAX_SERVER_LIMIT=$(((MEMORY_LIMIT_IN_BYTES/1024/1024 - 30) / 7))
+  HTTPD_SERVER_LIMIT=$((MAX_SERVER_LIMIT > 256 ? 256 : MAX_SERVER_LIMIT))
```
bparees: you need to enforce a minimum too, right?  otherwise you could set MAX_SERVER_LIMIT to zero here.

mnagy: True, good catch :)

mnagy:
PTAL

[test]

mfojtik:
@mnagy this should be the same as https://github.com/openshift/s2i-php/blob/master/5.6/s2i/bin/run#L26

mnagy:
OK, I'll try to unify this a bit, add additional settings and maybe make sure the code is more inline with php.

mnagy:
Unified with PHP, added HTTPD_START_SERVERS env var. PTAL

5.16/s2i/bin/run
```diff
@@ -17,4 +17,37 @@ PerlInitHandler Apache2::Reload
...
+    StartServers          ${HTTPD_START_SERVERS}
+    MinSpareServers       ${HTTPD_START_SERVERS}
+    MaxSpareServers       ${HTTPD_MAX_SPARE_SERVERS}
+    # The MaxRequestWorkers directive sets the limit on the number of simultaneous requests that will be served.
+    # The default value, when no cgroup limits are set is 256.
+    MaxRequestWorkers     ${HTTPD_MAX_REQUEST_WORKERS}
+    ServerLimit           ${HTTPD_MAX_REQUEST_WORKERS}
+    MaxRequestsPerChild   4000
+    MaxKeepAliveRequests  100
+</IfModule>
```
bparees: i think this would be better as a template file in the image that we run envsubst on, like we do in other images.

that way a user can override the file in the image if they extend our image, and not worry about our run script overwriting their override.

mnagy: They can always add another file (e.g. `/opt/app-root/etc/httpd.d/10-my-overrwrite.conf`) that sets different values. I like the fact that you can see directly in the run script what we're doing.

In any case, it would be best if this was all done in a (for now hypothetical) httpd base image.

bparees: can they unset a value you set here, by doing that?

this isn't how we do it for any other config value we set from env variables.

it also means if someone overrides the run script, they have to retain this block instead of just an envsubst line.

mnagy: They can overwrite any of the settings and every one of them has a default anyway.

I'll buy it on the run script override and consistency, so I'll update it.

bparees: overwrite isn't the same as unsetting.  if i want to get back to the default behavior, i'd now need to look up what that default setting is, and explicitly set it as my override value.

mnagy:
PTAL

5.16/s2i/bin/run
```diff
@@ -17,4 +17,24 @@ PerlInitHandler Apache2::Reload
...
+
+# If there is no server limit specified, try to guess the best value
+if [ -z "${HTTPD_MAX_REQUEST_WORKERS:-}" ]; then
+  MAX_SERVER_LIMIT=$(((MEMORY_LIMIT_IN_BYTES/1024/1024 - 30) / 7))
+  MAX_SERVER_LIMIT=$((MAX_SERVER_LIMIT > 0 ? MAX_SERVER_LIMIT : 1))
+  export HTTPD_MAX_REQUEST_WORKERS=$((MAX_SERVER_LIMIT > 256 ? 256 : MAX_SERVER_LIMIT))
+fi
+
+export HTTPD_START_SERVERS=${HTTPD_START_SERVERS:-8}
+export HTTPD_MAX_SPARE_SERVERS=$((HTTPD_START_SERVERS+10))
```
bparees: needs to check if it's set, so you don't override a user value?

mnagy: This is intentional so we don't expose yet another setting that needs documentation. I've used this from Michal's PHP PR.

bparees: ok

bparees:
lgtm

PR #84, opened 2016-05-06
don't pass an empty --env parameter to s2i

bparees:


ppisar:
Can you explain what was wrong with s2i --env=''?

This commit breaks the 4th psgi-variables test for me on Fedora 23 (source-to-image-1.0.3-2.fc23.x86_64) because the two --env arguments (--env=PSGI_FILE=./application2.psgi --env=PSGI_URI_PATH=/path) are not cumulative. The latter one overrides the former one, so the assemble script does not get PSGI_FILE variable.

I'd like to revert this commit completely. If you come with good excuse why --env='' is wrong, I will rewrite it so that it does pass --env option without arguments.

ppisar:
Now I can see. source-to-image-1.0.9-2.el7 in RHEL-7 changes --env handling. It would be good to upgrade the package in Fedora. Or I can augment the test to support both old and new s2i tool.

bparees:
@ppisar i'll see about getting the package updated.

@tdawson are you the one managing the fedora and rhel s2i packages?

ppisar:
Only Fedora 23 has the old source-to-image tool. I can understand if they
did not want to rebase it. In the mean I adjusted the tests to work with both
version. I will send a pull request if necessary.

PR #86, opened 2016-05-17
ensure /opt/app-root has proper permissions

bparees:


mnagy:
Why is this a good idea?

bparees:
to ensure the container user (group root) can actually write to those paths, which was the intent of the chown anyway.  But we ran into an issue when images are built on OSBS where the directory does not have group-write permission. 

it's needed because when we setup NSS we write (during startup) the password file to /opt/app-root/etc.  And i'm trying to make it consistent across all our images, even if we don't always write there.

bparees:
@mnagy see: https://bugzilla.redhat.com/show_bug.cgi?id=1336263

ppisar:
I must confess I'm loosing track of a security model in the containters. Do we
still pretend there is possibility to run processes with different UID/GID in
one container? If no, then why not simply grant all rights for the owner (or
even everybody) to everything in the /opt/app-root?

Now you start to accumulate various chmods and chowns on various files.
I wonder if somebody will remember their purpose after a year.

bparees:
> I must confess I'm loosing track of a security model in the containters. Do we still pretend there is possibility to run processes with different UID/GID in one container?

no

>  If no, then why not simply grant all rights for the owner (or even everybody) to everything in the /opt/app-root?

granting to "owner" is useless since the image can run as different uids, so the container uid will not necessary be the owner of the file.  That's why we ensure that files the container may need to access are

1) owned by the root group
2) g+rwx

because the container uid is always a member of the root group.

this this pull is granting the container uid (via the root group) all permissions on everything in /opt/app-root at the time the image is built.  fix-permissions does the same for things added to that folder during the assemble process.

bparees:
@mnagy does that make sense?

mnagy:
@bparees when you run the resulting image, will the process have right to write to /opt/app-root? Or is it just writable during the build?

bparees:
It will have access because the container uid is a member of the root group
and the directory has root group write permission.

Ben Parees | OpenShift

mnagy:
Isn't it not dangerous then?

mnagy:
And is the write permission necessary for the image to run?

bparees:
> And is the write permission necessary for the image to run?

@mnagy please read my original response to your questions on this PR:
https://github.com/openshift/s2i-perl/pull/86#issuecomment-219712190

and the bug i referenced that originally lead to this PR.

As for whether it's dangerous, yes if someone exploits your application such that they can perform operations as the container uid, this gives them elevated permissions.  We already accepted that when we introduced fix-permissions which makes the entire app directory writable so we could enable hot deployment.

the NSS passwd file has to be created during container startup because it has to be written with the current container uid, which is only known during container startup, so the file is always going to be owned/writable by the current container uid, there's no way around that.

mnagy:
LGTM, but it is horrible.

bparees:
Constructive suggestions appreciated.

Ben Parees | OpenShift

mnagy:
Sorry, I did not indend to criticize, but I'm a bit frustrated that we do not have an option for a more elegant solution..

PR #87, opened 2016-05-26
Remove IfModule check

mnagy:
Since we're processing our custom configuration before reading system
config files, the MPM prefork module is not loaded at the time of the
evaluation. Since this image will always run with Prefork (threaded
workers don't work with Perl), we can always set the settings, not
needing to check for the module to be loaded.

Fixes RHBZ https://bugzilla.redhat.com/show_bug.cgi?id=1339514

@bparees PTAL


bparees:
lgtm, but test issues.
[test]

mnagy:
I can reproduce it locally. I suspect a bug in tests, but not sure, investigating..

mnagy:
[test]

bparees:
[test]

bparees:
passed. merging.

PR #88, opened 2016-05-27
Make PSGI_FILE test more readable

mnagy:
@ppisar PTAL. This is a cosmetic change, but I had real troubles reading and parsing the line in my head, I am hoping this is much simpler now. I found this while trying to hunt a test failure which I'll fix in an additional update to this PR, but I want to make it fail first so I can see the difference.

[test]

### DIFF
--- a/5.20/s2i/bin/assemble
+++ b/5.20/s2i/bin/assemble
@@ -27,7 +27,7 @@ fi
 # If PSGI_FILE does not exist, check if exactly one ./*.psgi file exists and
 # use that file.
 # If PSGI_URI_PATH variable has a value, use it as a location. Default is "/".
-if [ ! -n "${PSGI_FILE+true}" ]; then
+if [ ! -v PSGI_FILE ]; then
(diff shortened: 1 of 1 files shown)


mnagy:
Failed as expected. Pushed update fixing [test]s, they should now pass.

ppisar:
This patch looks good.

mnagy:
Turns out that I was using old source-to-image binary. The new one changed usage of --env so that it is used multiple times. However, there's apparently a bug in s2i that causes the environment variables not to be passed to the build container.

mnagy:
@bparees cc about the bug. Caused by https://github.com/openshift/source-to-image/pull/501, I'm working on a fix.

bparees:
thanks @mnagy

mnagy:
[test]

mnagy:
re[test]

bparees:
[test]

bparees:
passed tests.  merging.

PR #90, opened 2016-06-28
Separate --mirror argument from CPAN_MIRROR env var

dinhxuanvu:
Currently, users need to add '--mirror' option along with mirror URL
in CPAN_MIRROR env var in order to work correctly. However, that option
'--mirror' shouldn't be a part of CPAN_MIRROR as it's rather confusing
to add an option into an env var that is used for mirror URL only.

Signed-off-by: Vu Dinh <vdinh@redhat.com>

### DIFF
--- a/5.16/s2i/bin/assemble
+++ b/5.16/s2i/bin/assemble
@@ -14,6 +14,12 @@ fi
 
 export CPAN_MIRROR=${CPAN_MIRROR:-""}
 
+MIRROR_ARGS=""
+
+if [ -n "$CPAN_MIRROR" ]; then
+  MIRROR_ARGS="--mirror $CPAN_MIRROR"
+fi
+
 # Don't test installed Perl modules by default
 if [ "${ENABLE_CPAN_TEST}" = true ]; then
   export ENABLE_CPAN_TEST=""
@@ -29,8 +35,8 @@ if [ -f "cpanfile" ]; then
     perl - App::cpanminus
   CPANM="./perl5/bin/cpanm"
   echo "---> Installing modules from cpanfile ..."
-  perl $CPANM $CPAN_MIRROR $ENABLE_CPAN_TEST -l extlib Module::CoreList
-  perl -Iextlib/lib/perl5 $CPANM $CPAN_MIRROR $ENABLE_CPAN_TEST -l extlib --installdeps .
+  perl $CPANM $MIRROR_ARGS $ENABLE_CPAN_TEST -l extlib Module::CoreList
+  perl -Iextlib/lib/perl5 $CPANM $MIRROR_ARGS $ENABLE_CPAN_TEST -l extlib --installdeps .
 else
   echo "---> No cpanfile found, nothing to install"
 fi
--- a/5.20/s2i/bin/assemble
+++ b/5.20/s2i/bin/assemble
@@ -14,6 +14,12 @@ fi
 
 export CPAN_MIRROR=${CPAN_MIRROR:-""}
 
+MIRROR_ARGS=""
+
+if [ -n "$CPAN_MIRROR" ]; then
+  MIRROR_ARGS="--mirror $CPAN_MIRROR"
+fi
+
 # Don't test installed Perl modules by default
 if [ "${ENABLE_CPAN_TEST}" = true ]; then
   export ENABLE_CPAN_TEST=""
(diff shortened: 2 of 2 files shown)


dinhxuanvu:
@bparees Please review. Thanks.

5.16/s2i/bin/assemble
```diff
@@ -29,8 +29,13 @@ if [ -f "cpanfile" ]; then
     perl - App::cpanminus
   CPANM="./perl5/bin/cpanm"
   echo "---> Installing modules from cpanfile ..."
-  perl $CPANM $CPAN_MIRROR $ENABLE_CPAN_TEST -l extlib Module::CoreList
-  perl -Iextlib/lib/perl5 $CPANM $CPAN_MIRROR $ENABLE_CPAN_TEST -l extlib --installdeps .
+  if [ -n "$CPAN_MIRROR" ]; then
```
bparees: needs to handle the case where someone was already using the old form and set their CPAN_MIRROR value to "--mirror somehost.com"

dinhxuanvu: In the doc, we do specifically say that CPAM_MIRROR should be for URL only. So, I don't think anyone would add the '--mirror' option. I only added the '--mirror' option only because that was how we did back in V2 which is bad in the first place.

bparees: but anyone who managed to get it working must have figured out on their own that they needed to add --mirror and done so, right?

dinhxuanvu: Well, that's true but I don't think anyone actually uses the CPAN_MIRROR option just yet as normally downloading packages from cpan.org by default is working fine. If they have noticed the issue, they would have probably filed a bug for that given we say that CPAN_MIRROR is URL only. I would rather not to pollute the assemble code with too many unnecessary checkpoints just because we try to tolerate an issue from the past.
Anyway, just my opinion on that. If you still think the change is necessary, I will fix it. No big deal. Happy to do it :).

bparees: I suppose you're right that someone who tried to use it probably would have opened an issue, and i agree i don't want to carry the if-check forever because of this.  I guess we can take the chance.

5.16/s2i/bin/assemble
```diff
@@ -29,8 +29,13 @@ if [ -f "cpanfile" ]; then
...
+    perl $CPANM --mirror $CPAN_MIRROR $ENABLE_CPAN_TEST -l extlib Module::CoreList
+    perl -Iextlib/lib/perl5 $CPANM --mirror $CPAN_MIRROR $ENABLE_CPAN_TEST -l extlib --installdeps .
+  else
+    perl $CPANM $ENABLE_CPAN_TEST -l extlib Module::CoreList
+    perl -Iextlib/lib/perl5 $CPANM $ENABLE_CPAN_TEST -l extlib --installdeps .
```
bparees: instead of repeating most of these two blocks, a better approach is:

MIRROR_ARGS=""
if [ $MIRROR ]; then
  MIRROR_ARGS="--mirror ${MIRROR}"
fi
perl blah blah $MIRROR_ARGS blah blah

dinhxuanvu: Good point. Fix is on the way. Thanks.

dinhxuanvu:
@bparees Ready for second round. Thanks :)

5.16/s2i/bin/assemble
```diff
@@ -29,8 +35,8 @@ if [ -f "cpanfile" ]; then
     perl - App::cpanminus
   CPANM="./perl5/bin/cpanm"
   echo "---> Installing modules from cpanfile ..."
-  perl $CPANM $CPAN_MIRROR $ENABLE_CPAN_TEST -l extlib Module::CoreList
-  perl -Iextlib/lib/perl5 $CPANM $CPAN_MIRROR $ENABLE_CPAN_TEST -l extlib --installdeps .
+  perl $CPANM MIRROR_ARGS $ENABLE_CPAN_TEST -l extlib Module::CoreList
```
bparees: needs $

5.16/s2i/bin/assemble
```diff
@@ -29,8 +35,8 @@ if [ -f "cpanfile" ]; then
...
+  perl -Iextlib/lib/perl5 $CPANM MIRROR_ARGS $ENABLE_CPAN_TEST -l extlib --installdeps .
```
bparees: needs $

dinhxuanvu:
Yah, I just noticed that when you pointed out. Fixed now. Sorry about that.

bparees:
lgtm.  fyi we've disabled pushing to dockerhub this week, so you won't see a new image w/ this change until next week.  (plus as mentioned in your other PR, the rhel image will need to go through the normal publication release process).

dinhxuanvu:
@bparees Acknowledged that. Will keep an eye on it. Thanks for the help :100:

5.16/s2i/bin/assemble
ppisar: I agree with the commit. The variable is a URL and should not contain
cpanminus-specific options. Thank you for the work.

PR #94, opened 2016-07-14
check for empty mirror value correctly

bparees:


ppisar:
What's the point of this change? Both the expressions are equivalent. The new
one is actually more complicated, hence less understandable.

bparees:
@ppisar there is no point, that's why i closed it :)

I thought there was a difference in behavior but i was wrong.

PR #95, opened 2016-07-20
Handle uppercase HTTP_PROXY

mnagy:
Curl and cpanm ignore HTTP_PROXY and only use http_proxy. Set http_proxy
to HTTP_PROXY in case only the later is set.

Output error if curl fails.

### DIFF
--- a/5.16/s2i/bin/assemble
+++ b/5.16/s2i/bin/assemble
@@ -13,6 +13,10 @@ if [ -d ./cfg ]; then
   fi
 fi
 
+if [[ -n "${HTTP_PROXY:-}" && -z "${http_proxy:-}" ]]; then
+  export http_proxy=$HTTP_PROXY
+fi
+
 export CPAN_MIRROR=${CPAN_MIRROR:-""}
 
 MIRROR_ARGS=""
@@ -32,7 +36,7 @@ fi
 if [ -f "cpanfile" ]; then
   # Install cpanm
   echo "---> Installing cpanminus 1.7102 ..."
-  curl -skL https://raw.githubusercontent.com/miyagawa/cpanminus/1.7102/cpanm |\
+  curl -sSkL https://raw.githubusercontent.com/miyagawa/cpanminus/1.7102/cpanm |\
     perl - App::cpanminus
   CPANM="./perl5/bin/cpanm"
   echo "---> Installing modules from cpanfile ..."
--- a/5.20/s2i/bin/assemble
+++ b/5.20/s2i/bin/assemble
@@ -13,6 +13,10 @@ if [ -d ./cfg ]; then
   fi
 fi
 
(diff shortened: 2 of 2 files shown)


mnagy:
@bparees PTAL

ppisar:
On Wed, Jul 20, 2016 at 02:15:33AM -0700, Martin Nagy wrote:
> Curl and cpanm ignore HTTP_PROXY and only use http_proxy.

That's fine because the correct variable name is lower case. The lower-cased
name is recognized by all software I know.

> Set http_proxy to HTTP_PROXY in case only the later is set.
>
I'm against adding support for non-standard variant.

5.16/s2i/bin/assemble
```diff
@@ -13,6 +13,10 @@ if [ -d ./cfg ]; then
   fi
 fi
 
+if [[ -n "${HTTP_PROXY:-}" && -z "${http_proxy:-}" ]]; then
+  export http_proxy=$HTTP_PROXY
+fi
```
bparees: add a comment about why we're doing this.

and does curl support both "HTTPS_PROXY" and "https_proxy"?  if not, we should be doing the same thing for that variable.

bparees:
@ppisar users don't know, and shouldn't have to know, that the assemble script in this image is running a curl command under the covers.  Our API for "user wants to use a proxy" is that they set the standard HTTP_PROXY/http_proxy/whatever env variables.  So if a user sets one of them, we have sufficient information to understand their intent and seamlessly give them that experience since we know curl (for whatever reason) doesn't explicitly handle it.  I fail to see any harm to doing this.

as you can see here, while it is true that lowercase is the convention, there is confusion on the internet as to which one people should use:
http://superuser.com/questions/876100/https-proxy-vs-https-proxy

so we are helping users by being agnostic about it, again particularly since they don't actually know what the assemble script is doing under the covers.

ppisar:
On Wed, Jul 20, 2016 at 05:52:54AM -0700, Ben Parees wrote:
> @ppisar users don't know, and shouldn't have to know, that the assemble
> script in this image is running a curl command under the covers.  Our API
> for "user wants to use a proxy" is that they set the standard
> HTTP_PROXY/http_proxy/whatever env variables.

I don't know your API. If you decided to support non-standard HTTP_PROXY, then
probably you should do the conversion to http_proxy on your level. Because all
the other universe uses http_proxy. Otherwise you will have to patch
everything you deliver to support HTTP_PROXY (e.g. this pull request).

I consider your approach nonscalable. But if you feel brave enough to
redefine universe, let continue.

bparees:
We aren't converting, we are amending to say "we will also tolerate this value because it's obvious what you meant and you shouldn't have to know that assemble uses a tool that doesn't tolerate that value even though many linux other tools do tolerate that value"

PR #98, opened 2016-08-10
add jenkins badge

PI-Victor:
@hhorak cc

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 Perl Docker images
 ==================
 
+[![Build Status](https://ci.openshift.redhat.com/jenkins/buildStatus/icon?job=sti-perl)](https://ci.openshift.redhat.com/jenkins/job/sti-perl)
+
 This repository contains the source for building various versions of
 the Perl application as a reproducible Docker image using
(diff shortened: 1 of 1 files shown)


bparees:
This (and the other ones) should not be merged. That job is run to test PRs, it doesn't reflect the state of master, it reflects the state of the last pr that was tested.

hhorak:
I already merged mongodb one (not others) before reading the comment above. So what we would need is to have a separate job for testing master, right? (not sure about Jenkins details, whether we can distinguish it within one job, common for PR and master)

bparees:
If you want a status badge, yes. Or the badge could be based on our push
images job which would indicate we've successfully built and pushed master
of all the repos, except that job doesn't build or push the latest rhel
images, so it's not a full source of truth. Also it isn't guaranteed to
have run on master yet since it runs hourly or so, so the badge could
indicate success when things were broken.

Overall I don't see much point in status badges like this.

Ben Parees | OpenShift

PI-Victor:
@bparees i've reverted this change now and submitted a PR, in my defense i did check the details on the jenkins page twice and it seemed ok with me, but i was clearly wrong.

PR #99, opened 2016-09-07
Support source-to-image-1.0.3

ppisar:
1.0.3 s2i tool does not support multiple --env arguments as introduced
in:

commit 08ad9bbaec9fd2f42addff74ee901b7005ecbd80
Author: Ben Parees bparees@redhat.com
Date:   Fri May 6 15:34:15 2016 -0400

```
don't pass an empty --env parameter to s2i
```

This patch adds a compatibility layer to tests to work with 1.0.3 again.

https://github.com/sclorg/s2i-perl-container/pull/84

### DIFF
--- a/5.20/test/run
+++ b/5.20/test/run
@@ -204,13 +204,32 @@ test_application() {
 # Second argument is function that expects running application. The function
 # must return (or terminate with) non-zero value to report an failure,
 # 0 otherwise.
-# Third argument is s2i --env argument value.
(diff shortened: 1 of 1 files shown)


bparees:
this is a lot of overhead to support an outdated version of s2i.  why don't we just update the s2i package?

ppisar:
Merge in pull request #103 with rh-perl524 support.

PR #103, opened 2016-10-27
rh-perl524 support

ppisar:
This patchset adds support for rh-perl524 collection. It adds support for source-to-image-1.0.3 as discussed on https://github.com/sclorg/s2i-perl-container/pull/99 to work on Fedora 23. Then it adds support for rh-perl524 collection and finally it disables the collection in the top-level Makefile because CentOS is still missing the rh-perl524 packages (not to break automated CentOS tests). The last commit will be reverted once the collection appears in the CentOS.


hhorak:
No changes from 5.20, except cpanminus is now available as RPM, which is nice. So LGTM. Thanks!

PR #106, opened 2016-11-10
Fix using subscription in container

omron93:
Enabling host repositories didn't work with docker 1.10 and newer inside OpenStack. It was fixed in latest rhel7 base release, so using 'rhel7' base image is required for RHEL CI. Postgresql PR - sclorg/postgresql-container#150

Also to enable repositories inside container running 'yum' first is required - see https://access.redhat.com/solutions/1443553 . Similar to sclorg/s2i-base-container#99 .

@hhorak @bparees Please take a look and merge.

### DIFF
--- a/5.16/Dockerfile.rhel7
+++ b/5.16/Dockerfile.rhel7
@@ -20,7 +20,10 @@ LABEL BZComponent="openshift-sti-perl-docker" \
       Architecture="x86_64"
 
 # TODO: Cleanup cpanp cache after cpanminus is installed?
-RUN yum-config-manager --enable rhel-server-rhscl-7-rpms && \
+# To use subscription inside container yum command has to be run first (before yum-config-manager)
+# https://access.redhat.com/solutions/1443553
+RUN yum repolist > /dev/null && \
+    yum-config-manager --enable rhel-server-rhscl-7-rpms && \
     yum-config-manager --enable rhel-7-server-optional-rpms && \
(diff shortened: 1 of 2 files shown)


ppisar:
On Thu, Nov 10, 2016 at 03:25:03AM -0800, Marek Skalický wrote:
> Enabling host repositories didn't work with docker 1.10 and newer inside
> OpenStack. It was fixed in latest rhel7 base release, so using 'rhel7' base
> image is required for RHEL CI. Postgresql PR
> - sclorg/postgresql-container#150
> 
> Also to enable repositories inside container running 'yum' first is required
> - see https://access.redhat.com/solutions/1443553 . Similar to
> sclorg/s2i-base-container#99 .
> 
Could you please write this into the commit message? The commit should explain
why you did the change.

Without any explanation in the code or in the commit message I would revert
this change because on the first sight, the "yum repolist" is absolutly
useless.

Moreover is this a work-around for temporary bug or is this a permament
solution?

-- Petr

omron93:
@ppisar Is it better now?

> Moreover is this a work-around for temporary bug or is this a permament solution?

This is how using subscription in rhel containers officially works.

From https://access.redhat.com/solutions/1443553:
> NOTE: Until the first yum command is run, /etc/yum.repos.d/redhat.repo contains no repositories, so yum-config-manager will not enable/disable anything.

bparees:
even better, put a comment in the dockerfile explaining why it's being done.

hhorak:
Makes sense, LGTM.

PR #107, opened 2016-11-11
Fix using subscription in container

hhorak:
To use subscription inside container yum command has to be run first (before yum-config-manager)


hhorak:
This is the same change as done in PR#106, which won't be tested now anyway, so merging.

PR #109, opened 2016-11-11
New labels for OSBS

hhorak:
Labels now should follow spec at https://github.com/projectatomic/ContainerApplicationGenericLabels/blob/master/vendor/redhat/labels.md


ppisar:
In line with the new label convetions
<https://github.com/projectatomic/ContainerApplicationGenericLabels/blob/master/vendor/redhat/labels.md#naming-and-versioning-related-labels>.
This is backward incompatible change.

hhorak:
> This is backward incompatible change.

Yes, but there are reason why I believe we can still do this change:
* the change is only in capitalization (except BZComponent)
* osbs authors proposed exactly this change (no problem for internal build service)
* we didn't document this as something what is "part of API" and people should depend on

PR #111, opened 2016-12-07
Pass image ID for testing

omron93:
When running test/run use image ID instead of IMAGE_NAME. This allows parallel testing of images (for example parallel testing of PRs in Jenkins).

This is similar PR as for sclorg/mariadb-container#14. @hhorak Please take a look.

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -13,6 +13,9 @@
 # OPENSHIFT_NAMESPACES - Which of available versions (subdirectories) should be
 #       put into openshift/ namespace.
 
+set -o errexit
+set -o pipefail
+
 OS=${1-$OS}
 VERSION=${2-$VERSION}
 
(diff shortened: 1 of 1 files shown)


ppisar:
On Wed, Dec 07, 2016 at 07:57:52AM -0800, Marek Skalický wrote:
> When running test/run use image ID instead of IMAGE_NAME. This allows
> parallel testing of images (for example parallel testing of PRs in Jenkins).
> 
So you execute test/run like this:

-    IMAGE_NAME=${IMAGE_NAME} test/run
+    IMAGE_NAME=${IMAGE_ID} test/run

But 5.24/test/run does use IMAGE_ID at all. I think you should modify
*/test/run scripts too.

omron93:
> + IMAGE_NAME=${IMAGE_ID} test/run
>
> But 5.24/test/run does use IMAGE_ID at all. I think you should modify */test/run scripts too.

@ppisar Value of IMAGE_NAME in `test/run` is not a tag of the image, but image ID. But I might not see something...

Problem is, that `s2i build` is not handling image identifier as docker do. So with `s2i` it is not possible to use image ID instead of its name. So this patch have to be reworked.

omron93:
Approach used in this PR is not possible for s2i based images (running `s2i build` in tests fails - using image ID instead of tag is not supported). So am closing this PR (and similar PRs for other container images) to keep building scripts same s2i based and non-s2i based images.

As mentioned in sclorg/postgresql-container#155, different approach could be used, so I will try to prepare something. But it would be helpful to have shared build scripts as mentioned in sclorg/postgresql-container#138.

ppisar:
On Thu, Dec 08, 2016 at 12:53:11AM -0800, Marek Skalický wrote:
> > + IMAGE_NAME=${IMAGE_ID} test/run
> >
> > But 5.24/test/run does use IMAGE_ID at all. I think you should modify */test/run scripts too.
> 
> @ppisar Value of IMAGE_NAME in `test/run` is not a tag of the image, but
> image ID. But I might not see something...
>
I see. I overlooked that you still passes IMAGE_NAME variable. Then please
ignore my comment.

PR #117, opened 2017-05-26
Update Dockerfiles

omron93:
Provide same labels for all base image variants (CentOS, RHEL)
 - follow recomended labels from Project Atomic Container best practices
 - add summary and description labels - use ENV for simple using of same value in more descriptive labels
 - remove architecture label for RHEL based images (this label is set in RHEL image - no need to define it again)
                                           
(don't use separate instruction for each label)

@hhorak @praiskup @pkubatrh Please take a look and merge.


ppisar:
Why do you add Fedora specific labels (summary, description) into RHEL and
CentoOS images?

Why do you add com.redhat.component label to CentoOS images? There is nothing
like that in CentOS bug tracking system. <https://bugzilla.redhat.com/> is
used for RHEL-based images only. Everybody else uses
<https://github.com/openshift/sti-perl/>. Like you.

omron93:
> Why do you add Fedora specific labels (summary, description) into RHEL and CentoOS images? 

Sorry, it is Fedora requirements. But Project Atomic also has it - https://github.com/projectatomic/ContainerApplicationGenericLabels/blob/master/vendor/redhat/labels.md (also RH Container Catalog will use these labels)

> Why do you add com.redhat.component label to CentoOS images? There is nothing like that in CentOS bug tracking system. <https://bugzilla.redhat.com/> is used for RHEL-based images only.

Fedora has this in its guidelines - https://fedoraproject.org/wiki/Container:Guidelines#LABELS
I am not sure about it, but I think we recommended users to report bugs in centos based images also to RH bugzilla - to have bugs in one place. (same for CentOS rebuilds of Software Collections. @hhorak Where the bugs should be reported?

> Everybody else uses <https://github.com/openshift/sti-perl/>. Like you.

@ppisar What does this means please?

ppisar:
On Fri, May 26, 2017 at 06:12:38AM -0700, Marek Skalický wrote:
> > Why do you add Fedora specific labels (summary, description) into RHEL and CentoOS images? 
> 
> Sorry, it is Fedora requirements.

Dockerfile is for CentOS.
Dockerfile.rhel is for RHEL.

None of them is for Fedora.

> But Project Atomic also has it
> - https://github.com/projectatomic/ContainerApplicationGenericLabels/blob/master/vendor/redhat/labels.md
> (also RH Container Catalog will use these labels)
>
So it's a RHEL requirement. Ok.

> > Why do you add com.redhat.component label to CentoOS images? There is
> > nothing like that in CentOS bug tracking system.
> > <https://bugzilla.redhat.com/> is used for RHEL-based images only.
> 
> Fedora has this in its guidelines
> - https://fedoraproject.org/wiki/Container:Guidelines#LABELS I am not sure
> about it, but I think we recommended users to report bugs in centos based
> images also to RH bugzilla - to have bugs in one place. (same for CentOS
> rebuilds of Software Collections. @hhorak Where the bugs should be reported?
>
If it so, then it's fine.

> > Everybody else uses <https://github.com/openshift/sti-perl/>. Like you.
> 
> @ppisar What does this means please?
> 
This means that you sent this to Github. Not Red Hat's Bugzilla. Thus
I dispute usefullness or pointing users to Bugzilla if actually all the
development happens elsewhere.

Now, when it turned out this is not about Fedora, but about RHEL and CentOS,
I'm fine changing RHEL and CentOS images as you proposed in the patch.

omron93:
> None of them is for Fedora.

Sorry. I am doing this update for all sclorg/ images and a lot of them contains also Fedora dockerfiles. And the message was copy&pasted :-(
I've removed mentioning of Fedora from this PR...

hhorak:
The PR looks good to me now, @ppisar are you find with that now or are there still some unresolved issues?

ppisar:
I'm fine with these patches.

pkubatrh:
Merging, thanks.

PR #122, opened 2017-09-20
Enable rh-nodejs6, so users can use npm for installing JavaScript stuff

hhorak:
Similarly to https://github.com/sclorg/s2i-python-container/pull/209


ppisar:
It's fine. You can merge it.

In ideal world, I'd like to see this in the s2i-base image, so that you don't
have edit each layered image, but I guess this out of scope of this feature
now.

PR #128, opened 2017-12-10
Add Openshift tests

hhorak:
This change requires the following change to be merged in the common submodule: https://github.com/sclorg/container-common-scripts/pull/46
After that, the test will try deploy sample application and dancer example application in OpenShift environment.


5.20/test/run-openshift
```diff
@@ -0,0 +1,45 @@
...
+  ct_os_wait_pod_ready "${service_name}" 60
+  local ip=$(ct_os_get_service_ip "${service_name}")
+  ct_test_response "http://${ip}:8080" 200 "${expected_output}"
+
+  ct_os_delete_project
+}
+
+ct_os_cluster_up
+# TODO: Can we make the build against examples inside the same PR?
+ct_os_test_s2i_app ${IMAGE_NAME} "https://github.com/sclorg/s2i-perl-container.git" 5.24/test/sample-test-app "Everything is OK"
```
ppisar: This test adds a dependency on a resource provided by a third-party on the Internet. This is generally bad idea.

hhorak: `oc new-app` should work also with a local path, so let's try that variant.

hhorak: Hm, I'm trying to find out how to do it and it looks not easy actually. @pvalena, do you know how to use the local application (files in a local dir) and a local image in the `oc new-app`?

pvalena: If you want to use local files, you have to specify path to local directory, instead of url. Otherwise it's completely same.
From their examples:
```bash
$ oc new-app openshift/ruby-20-centos7:latest~/home/user/code/my-ruby-app
```
https://docs.openshift.org/latest/dev_guide/application_lifecycle/new_app.html#dev-guide-new-app

pvalena: Regarding the url to sample repo, that's actually [distributed](https://github.com/openshift/origin/blob/master/examples/quickstarts/dancer-mysql.json#L13) by default in OpenShift quickstarts and we're AFAIK supposed to maintain or fix it in case of need.

pvalena:
More about [quickstarts](https://docs.openshift.com/container-platform/3.7/dev_guide/migrating_applications/quickstart_examples.html).

5.20/test/run-openshift
pvalena: More about [quickstarts](https://docs.openshift.com/container-platform/3.7/dev_guide/migrating_applications/quickstart_examples.html).

pkubatrh: >If you want to use local files, you have to specify path to local directory, instead of url. 

Be aware that if you provide a path that points to a git repository and that git repository has an `origin` remote Openshift will just take the code from the remote instead of the local files.

pkubatrh: Additionally after running `oc new-app` with a local path as an argument (not containing a git repo) I had to actually run `oc start-build --from-dir` manually as Openshift only created the buildconfig.

hhorak:
[test-openshift]

pkubatrh:
Openshift tests ran without an issue for me locally. Possibly just needs a bigger timeout for the build phase.

5.20/test/run-openshift
```diff
@@ -0,0 +1,44 @@
...
+
+  ct_os_new_project
+  ct_os_upload_image "${image_name}" "perl:$VERSION"
+
+  ct_os_deploy_template_image ${THISDIR}/../../examples/templates/sample-test-app.json \
+                              NAMESPACE="$(oc project -q)"
+
+  ct_os_wait_pod_ready "${service_name}" 60
+  local ip=$(ct_os_get_service_ip "${service_name}")
+  ct_test_response "http://${ip}:8080" 200 "${expected_output}"
```
pkubatrh: This causes a lot of log polution when -x is set since the response is stored in a variable inside the library function.

hhorak:
[test-openshift]

5.20/test/run-openshift
hhorak: Should be fixed now.

hhorak:
[test-openshift]

hhorak:
Tests pass now, I guess we can merge. @ppisar, any last comments?

5.20/test/run-openshift
hhorak: As for the original comment, testing third-party components from the Internet is actually what we should do here, I believe. IOW we need to verify that whatever is used by users (which are those third-party components, templates and quickstarts), that also works with the updated image..

ppisar:
It looks good now.

5.20/test/run-openshift
```diff
@@ -0,0 +1,44 @@
...
+
+function test_sample_template() {
+  local image_name=${1:-centos/perl-524-centos7}
+  local image_name_no_namespace=${image_name##*/}
+  local service_name=${image_name_no_namespace}
+
+  ct_os_new_project
+  ct_os_upload_image "${image_name}" "perl:$VERSION"
+
+  ct_os_deploy_template_image ${THISDIR}/../../examples/templates/sample-test-app.json \
```
pvalena: I'd stick with symlinks, as other containers use them instead of `../../path/to/template`, as it's more error prone and flexible, in case you need to replace the template f.e. with some experimental one.

hhorak: OK, I'll fix this one.

5.24/test/run-openshift
```diff
@@ -0,0 +1,44 @@
...
+
+function test_sample_template() {
+  local image_name=${1:-centos/perl-524-centos7}
+  local image_name_no_namespace=${image_name##*/}
+  local service_name=${image_name_no_namespace}
+
+  ct_os_new_project
+  ct_os_upload_image "${image_name}" "perl:$VERSION"
+
+  ct_os_deploy_template_image ${THISDIR}/../../examples/templates/sample-test-app.json \
```
pvalena: Same fix here.

pvalena:
Otherwise LGTM.

hhorak:
[test-openshift]

PR #130, opened 2017-12-18
Dockerfile: put variables inside curly braces

pkubatrh:
since OSBS had trouble expanding some during build


5.20/Dockerfile
```diff
@@ -23,15 +23,15 @@ LABEL summary="$SUMMARY" \
       io.k8s.description="$DESCRIPTION" \
       io.k8s.display-name="Apache 2.4 with mod_perl/$PERL_VERSION" \
       io.openshift.expose-services="8080:http" \
-      io.openshift.tags="builder,$NAME,$NAME$PERL_SHORT_VER" \
+      io.openshift.tags="builder,$NAME,${NAME}${PERL_SHORT_VER}" \
```
ppisar: The first $NAME is not parenthesized. This flaw is in all Dockerfile files.

pkubatrh: Fixed now, thanks!

ppisar:
Now the patch is good.

PR #140, opened 2018-05-14
Add version 5.26

pkubatrh:
The centos build will fail for now as the `rh-perl526-perl-Apache-Reload` is yet to be rebuilt, but the code can be ready in this PR until it is available.


ppisar:
Thank you for the pull request, but I have already made all necessary changes. But not yet pushed because of CentOS lacking the builds.

hhorak:
[test]

hhorak:
[test-openshift]

hhorak:
It seems like the openshift tests only fail because there is no new app in the master yet, which should be fixed after merging. So, merging :) THANKS!

PR #141, opened 2018-05-14
Deprecate version 5.20

pkubatrh:


omron93:
Thanks.

ppisar:
The Makefile should move 5.20 from VERSIONS variable to OPENSHIFT_NAMESPACES
variable. The patch removes it instead. OpenShift wants to offer deprecated
images to its users.

pkubatrh:
To be honest I do not think the `OPENSHIFT_NAMESPACES` variable is even used for anything anymore

omron93:
`OPENSHIFT_NAMESPACES` was used to pick images which should have `openshift/` namespace. Later images now have `centos/` namespace.

PR #147, opened 2018-07-09
imagestream definition for rhel based perl 5.26 on ppc64le

jeffdyoung:
Adding rhel7 based perl 5.26 imagestream that will support ppc64le.


ppisar:
No objections.

PR #149, opened 2018-09-17
rename imagestream from perl_ppc64le to perl

jeffdyoung:
rename imagestream from perl_ppc64le to perl


ppisar:
No issues from the Perl point of view.

hhorak:
LGTM, thanks.

PR #159, opened 2019-02-12
5.24: Install all core modules

ppisar:
Some core modules do not live on CPAN and cannot be installed in an
assemble phase. E.g. Devel::Peek.

This patch preinstalls rh-perl524-perl-core RPM package that ensures
that all core modules are available in the assemble phase.

This is not an issue for 5.26 where the same guarantee is implemented
by explicitly installed rh-perl526-perl RPM package.


ppisar:
[test-openshift]

hhorak:
[test-openshift]

hhorak:
[test]

hhorak:
LGTM, just re-running the test suite since centos one did not start from unknown reason.

PR #160, opened 2019-03-05
remove help.md from 5.26-mod_fcgid

pkubatrh:
as it has the same content as README.md and is not kept as a symlink in other versions


ppisar:
I'm fine with removing the file.

pkubatrh:
Thanks for review, merging.

PR #165, opened 2019-06-26
Fix documentation for environment variables

torsava:
Corect the directory name: .sti -> .s2i

### DIFF
--- a/5.24/README.md
+++ b/5.24/README.md
@@ -50,7 +50,7 @@ $ curl 127.0.0.1:8080
 Environment variables
 ---------------------
 
-To set environment variables, you can place them as a key value pair into a `.sti/environment`
+To set environment variables, you can place them as a key value pair into a `.s2i/environment`
 file inside your source code repository.
 
 * **ENABLE_CPAN_TEST**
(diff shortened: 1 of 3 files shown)


ppisar:
May I know what code reads .sti/environment? I cannot find it either in
sti-perl or in sti-base repository. Is that hardcoded into s2i tool? What
version? I'd like verify that this change is correct.

ppisar:
I find this code in source-to-image-226afa1319c3498f47b974ec8ceb36526341a19c that is package in Fedora 29:

```
// GetEnvironment gets the .s2i/environment file located in the sources and
// parse it into EnvironmentList.
func GetEnvironment(config *api.Config) (api.EnvironmentList, error) {
    envPath := filepath.Join(config.WorkingDir, api.Source, ".s2i", api.Environment)
    if _, err := os.Stat(envPath); os.IsNotExist(err) {
        // TODO: Remove this when the '.sti/environment' is deprecated.
        envPath = filepath.Join(config.WorkingDir, api.Source, ".sti", api.Environment)
        if _, err := os.Stat(envPath); os.IsNotExist(err) {
            return nil, errors.New("no environment file found in application sources")
        }
        glog.Info("DEPRECATED: Use .s2i/environment instead of .sti/environment")
    }
```

According to the code "sti" is deprecated. "s2i" is the modern spelling. Do you have a different source? It looks like your patch is wrong.

ppisar:
Actually your patch is in line with the code. I will merge it.

PR #166, opened 2019-07-14
remove versions that are EOL

gabemontero:
per request from @siamaksade from OpenShift PM, this PR removes EOF version from the imagestream file

@hhorak @bparees


ppisar:
It looks good.

PR #168, opened 2019-09-03
add license terms for rhel based images

pkubatrh:


ppisar:
The patch looks good.

PR #169, opened 2019-09-20
Fix permissions for all files given in s2i

phracek:

Adds fix-permissions right after `mv` command. `cp` command also has parameter `-a`.

Reference issue:
https://github.com/sclorg/container-common-scripts/issues/119

### DIFF
--- a/5.24/s2i/bin/assemble
+++ b/5.24/s2i/bin/assemble
@@ -6,10 +6,13 @@ shopt -s dotglob
 echo "---> Installing application source ..."
 mv /tmp/src/* ./
 
+# Fix source directory permissions
+fix-permissions ./
+
 if [ -d ./cfg ]; then
   echo "---> Copying configuration files..."
   if [ "$(ls -A ./cfg/*.conf)" ]; then
-    cp -v ./cfg/*.conf /opt/app-root/etc/httpd.d/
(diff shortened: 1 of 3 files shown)


praiskup:
Why this change?  Can it be documented in the commit message?

pkubatrh:
Tied to https://github.com/sclorg/container-common-scripts/issues/119

But I agree it should get a comment in the commit message at least.

pkubatrh:
Please also add `-a` to the `cp` calls after the `fix-permission` script is done.

ppisar:
The assemble script creates files after your new location of fix-permissions
invocation. What permissions and group will they get? I believe they will
inherit them from s2i program. I don't believe this fix is correct.

Either set EGID and umask right after the new new invocation or keep the
fix-permissions also on the old location.

phracek:
@ppisar `fix-permissions` are always called in this container on directory `./`. I don't follow why it should be called twice.

ppisar:
Now the patch looks good. Only the commit message is strange. The message reads "Move" but the patch "adds". Also you are missing something between "command keep" words.

phracek:
I have updated the commit message.

pkubatrh:
[test][test-openshift]

pkubatrh:
[test-openshift]

pkubatrh:
Tests are green so merging, thanks!

PR #170, opened 2019-10-24
Add support for internal Nexus registry

phracek:
Add parameter `ct_build_s2i_npm_variables`
into `run_s2i_build` function.
It checks if internal NPM_REGISTRY is set and
CA file is present. Else use public NPM registry.

### DIFF
--- a/5.24/s2i/bin/assemble
+++ b/5.24/s2i/bin/assemble
@@ -29,6 +29,11 @@ if [ -n "$CPAN_MIRROR" ]; then
   MIRROR_ARGS="--mirror $CPAN_MIRROR"
 fi
 
+# Change the npm registry mirror if provided
+if [ -n "$NPM_MIRROR" ]; then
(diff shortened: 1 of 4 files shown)


phracek:
Reference https://github.com/sclorg/container-common-scripts/pull/127

phracek:
Thanks for the review. See the definition here https://github.com/sclorg/container-common-scripts/blob/master/test-lib.sh#L234

ppisar:
Then you need to rebase the "common" submodule within this pull request because now s2i-perl-conteiner master refers to common repository at f7fbbb3a9373affefbf0c17bbbc9ff082f7ece99 that does not know the ct_build_s2i_npm_variables() function.

Otherwise the commit is good.

phracek:
@ppisar Can you please look at it once more?

ppisar:
common submodule is rebased correctly. No objections.

PR #174, opened 2020-01-20
Add 5.30

ppisar:
This is an image with RHSCL-based Perl 5.30 with mod_perl.

It is excluded for all distributions because the collection has not
yet been released. The imagestreams were not ammended for the same reason.

There is also no 5.30-mod_fcgid because that one is based on the
RHEL 8 modules.

### DIFF
--- a/5.30/Dockerfile.rhel7
+++ b/5.30/Dockerfile.rhel7
@@ -1,0 +1,67 @@
+FROM rhscl/s2i-base-rhel7
+
+# This image provides a Perl 5.30 environment you can use to run your Perl applications.
+
+EXPOSE 8080
+
+# Image metadata
+ENV PERL_VERSION=5.30 \
+    PERL_SHORT_VER=530 \
+    NAME=perl
+
+ENV SUMMARY="Platform for building and running Perl $PERL_VERSION applications" \
+    DESCRIPTION="Perl $PERL_VERSION available as container is a base platform for \
+building and running various Perl $PERL_VERSION applications and frameworks. \
+Perl is a high-level programming language with roots in C, sed, awk and shell scripting. \
+Perl is good at handling processes and files, and is especially good at handling text. \
+Perl's hallmarks are practicality and efficiency. While it is used to do a lot of \
+different things, Perl's most common applications are system administration utilities \
+and web programming."
+
+LABEL summary="$SUMMARY" \
+      description="$DESCRIPTION" \
+      io.k8s.description="$DESCRIPTION" \
+      io.k8s.display-name="Apache 2.4 with mod_perl/$PERL_VERSION" \
+      io.openshift.expose-services="8080:http" \
+      io.openshift.tags="builder,${NAME},${NAME}${PERL_SHORT_VER}" \
+      io.openshift.s2i.scripts-url="image:///usr/libexec/s2i" \
+      io.s2i.scripts-url="image:///usr/libexec/s2i" \
+      name="rhscl/${NAME}-${PERL_SHORT_VER}-rhel7" \
+      com.redhat.component="rh-${NAME}${PERL_SHORT_VER}-container" \
+      version="$PERL_VERSION" \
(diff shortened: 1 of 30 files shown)


5.30/Dockerfile.rhel7
```diff
@@ -0,0 +1,67 @@
...
+      io.k8s.description="$DESCRIPTION" \
+      io.k8s.display-name="Apache 2.4 with mod_perl/$PERL_VERSION" \
+      io.openshift.expose-services="8080:http" \
+      io.openshift.tags="builder,${NAME},${NAME}${PERL_SHORT_VER}" \
+      io.openshift.s2i.scripts-url="image:///usr/libexec/s2i" \
+      io.s2i.scripts-url="image:///usr/libexec/s2i" \
+      name="rhscl/${NAME}-${PERL_SHORT_VER}-rhel7" \
+      com.redhat.component="rh-${NAME}${PERL_SHORT_VER}-container" \
+      version="$PERL_VERSION" \
+      com.redhat.license_terms="https://www.redhat.com/en/about/red-hat-end-user-license-agreements#rhel" \
```
phracek:
```suggestion
      com.redhat.license_terms="https://www.redhat.com/en/about/red-hat-end-user-license-agreements#UBI" \
```

5.30/README.md
```diff
@@ -0,0 +1,104 @@
...
+    This variable overrides location URI path that is handled path the PSGI
+    application. Default value is "/".
+
+
+See also
+--------
+Dockerfile and other sources are available on https://github.com/sclorg/s2i-perl-container.
+In that repository you also can find another versions of Python environment Dockerfiles.
+Dockerfile for CentOS is called `Dockerfile`, Dockerfile for RHEL7 is called `Dockerfile.rhel7`,
+for RHEL8 it's `Dockerfile.rhel8` and the Fedora Dockerfile is called Dockerfile.fedora.
```
phracek: I would remove RHEL8, because of s2i-perl-container does not have `Dockerfile.rhel8` at all.

Makefile
```diff
@@ -1,6 +1,6 @@
 # Variables are documented in common/build.sh.
 BASE_IMAGE_NAME = perl
-VERSIONS = 5.26 5.26-mod_fcgid
+VERSIONS = 5.30 5.26 5.26-mod_fcgid
```
phracek: I don't want to be a nitpicker, but does it make sense to have also 5.30-mod_fcig too?

5.30/Dockerfile.rhel7
pkubatrh: Is it going to be UBI? Just asking since the previous version is only released in the rhscl namespace (though rhel8 version is UBI).

Makefile
pkubatrh: I would say it does not make sense at this point, since the new version is being added for RHSCL, not RHEL8 as well.

5.30/Dockerfile.rhel7
phracek: Good point. I saw an s2i-ruby-container https://github.com/sclorg/s2i-ruby-container/pull/251 change and though that s2i-perl-container should use UBI too. A link would be nice where we can find which image is UBI and which is only RHSCL.

Makefile
phracek: Ok, make sense.

5.30/Dockerfile.rhel7
pkubatrh: https://access.redhat.com/containers/#/product/5c180b28bed8bd75a2c29a63

phracek: Thanks for the link.

5.30/README.md
ppisar: You are right. I removed it and rebased the pull request in bc9e1d40bb74cfea75c49202123865a2ca99a529 commit .

phracek:
Thanks for updating PR!. Running OpenShift test
[test-openshift]

PR #175, opened 2020-02-06
Deprecated Fedora 5.26 container

phracek:
5.26 version contains f28/s2i-base. This image does not exist anymore.
Switching to f29/s2i-base does not make sense because of Fedora 29 
contains `perl-5.28` version.

Therefore s2i-perl-container does not support 5.26 version in Fedora land.


pkubatrh:
```
docker run --rm -ti fedora:30 bash -c "dnf module list | grep perl"
Trying to pull docker.io/library/fedora:30...
Getting image source signatures
Copying blob b0fe058ffda2 done  
Copying config 89d6d6a7d5 done  
Writing manifest to image destination
Storing signatures
perl                 5.26           default [d], minimal                              Practical Extraction and Report Language                                       
perl-bootstrap       5.26                                                             Perl bootstrap module for bootrapping Perl module                              
perl                 5.26           default [d], minimal                              Practical Extraction and Report Language                                       
perl                 5.28           common, default [d], minimal                      Practical Extraction and Report Language                                       
perl                 5.30           common, default [d], minimal                      Practical Extraction and Report Language                                       
perl-bootstrap       5.26
```

phracek:
Really good catch!. Thank you. I totally forgot, that we have modules. I will update PR soon.

ppisar:
The module does not help you.

First, there is no module with mod_perl package that is needed for this container. Secondly, perl:5.26 support ended on 2019-12-01 <https://pdc.fedoraproject.org/rest_api/v1/component-branch-slas/?branch_type=module&global_component=perl&branch=5.26>. Do not use it. Lastly, for modularized packages, there is 5.26-mod_fcgi container. The non-mod_fcgi containers use mod_perl and are supposed not to use modules. That's to reflect the split between RHEL 7 and 8 and a switch from collections to modules.

In other words, this commit is perfect and I recommend merging it.

pkubatrh:
Thanks for the info @ppisar! I did not want to close it out-right as I suspected there might be some catch with the perl modules and turns out my hunch was right.

Just out of curiosity, is there something similar to 5.26-mod_fcgi (or any other version) in modular Fedora we could use for container builds, or is it RHEL8 only?

ppisar:
In modular Fedora there is only perl. Nothing else.

PR #178, opened 2020-06-22
Test the latest versions in the imagestreams

phracek:
This commit adds a test for checking the latest version in imagestreams.

### DIFF
--- a/5.26-mod_fcgid/test/run
+++ b/5.26-mod_fcgid/test/run
@@ -21,6 +21,7 @@ test_psgi_variables
 test_warningonstderr
 test_fcgi
 test_npm
+test_latest_imagestreams
 "
 
 source "${test_dir}/test-lib.sh"
@@ -353,6 +354,17 @@ test_npm() {
     check_result $?
 
     cleanup
+}
+
+test_latest_imagestreams() {
+  local result=1
+  info "Testing the latest version in imagestreams"
+  # Switch to root directory of a container
+  pushd "${test_dir}/../.." >/dev/null || return 1
+  ct_check_latest_imagestreams
+  result=$?
+  popd >/dev/null || return 1
+  check_result $result
 }
 
 function run_all_tests() {
--- a/5.26/test/run
+++ b/5.26/test/run
@@ -20,6 +20,7 @@ test_psgi
 test_psgi_variables
 test_warningonstderr
 test_npm
+test_latest_imagestreams
 "
 
 source "${test_dir}/test-lib.sh"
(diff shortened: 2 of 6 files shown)


ppisar:
It looks good except for:

It does not add a test into 5.26-mod_fcgid.
The second commit title "The the latest version in imagestreams" is garbled.
Shouldn't the first "The" be "Test"?

phracek:
@pkubatrh @hhorak Failure is expected. We don't have streams yet, I guess. Where can I clarify it s2-php-container already container version 5.30
@bparees Can you help me If I can add version 5.30 into imagestreams?

bparees:
> @bparees Can you help me If I can add version 5.30 into imagestreams?

@gabemontero can shepherd you through getting it added.  the steps are roughly:

1) add the tag to the imagestream in this repo and change the `latest` tag in the imagestream to point to the new one.
2) get openshift/library updated w/ the new tag (should happen automatically since that imagestream is already imported from this repo nightly, i believe)
3) get @gabemontero to update the openshift samples operator w/ the latest content from openshift/library

phracek:
@gabemontero PR for updating openshift/library is here https://github.com/openshift/library/pull/224
Please update also operator. Thank you.

phracek:
[test][test-openshift]

phracek:
Test failed on: 11:41:37 Step 1/14 : FROM registry.fedoraproject.org/f31/s2i-base:latest
11:41:37 Trying to pull repository registry.fedoraproject.org/f31/s2i-base ... 
11:41:37 received unexpected HTTP status: 503 Service Unavailable

Let's try one more time

[test][test-openshift]

phracek:
@ppisar I have added test for the latest imagestreams also for perl 5.26-mod_fcgi. Please have a look on it once more. Thanks.

ppisar:
Are you sure it's right to do that (call ct_check_latest_imagestreams), if
there is no imagestreams/perl-rhel8.json that would define what is the latest
stream on RHEL 8?

phracek:
> Are you sure it's right to do that (call ct_check_latest_imagestreams), if there is no imagestreams/perl-rhel8.json that would define what is the latest stream on RHEL 8?

Thanks for the question. As you can see in rhel8 build (https://gist.github.com/rhscl-automation/ebc07eab20e409c64a3753ecb233edca) we check all imagestreams, CentOS7, RHEL7, Fedora if exists. There is no such heuristic like checking RHEL7 imagestreams for RHEL7 only. But definitelly good idea.

phracek:
This should be easily implemented here https://github.com/sclorg/container-common-scripts/blob/master/check_imagestreams.py#L38

ppisar:
No objections from Perl point of view.

PR #179, opened 2020-07-13
Update imagestreams

yselkowitz:


ppisar:
No objections from point of view of Perl language and images.

phracek:
@yselkowitz Can you please update also perl-centos7.json imagestream? Why did you remove file  imagestreams/perl-rhel7-ppc64le.json

yselkowitz:
I don't see a perl-530-centos7 on docker.io right now, so wouldn't that be premature?

The arch-specific rhel7 files were removed because they are no longer used.  Now that the pre-multi-arch versions have been dropped, the main file is used for all RHEL 7 architectures:

https://github.com/openshift/library/blob/master/official.yaml#L127-L139

phracek:
Merging it. We need the latest version for https://github.com/openshift/library/pull/224

PR #180, opened 2020-07-16
Run all tests and reports failures at the end

phracek:
Run all tests and reports failures at the end.

The commit executes all tests and reports
at the end result in format like:

[PASSED] test_sample_test_app
[PASSED] test_binpath
[PASSED] test_psgi
[PASSED] test_psgi_variables
[PASSED] test_warningonstderr
[PASSED] test_npm

Tests for rhscl/perl-526-rhel7 succeeded.


phracek:
[test]

phracek:
[test-openshift]

ppisar:
I doubt it will work. the do_test() invocation sets test_name variable from its first argument and the variable is used in in run_s2i_build() to build an application image from the proper sources found in the same-named directory. Your patch removes the do_test() invocations and thus I believe the tests won't build the intended application images and will fail.

ppisar:
I got confused by renaming the test function. It looks good.

PR #182, opened 2020-07-28
imagestreams/perl-centos7: remove 5.30

vrutkovs:
Perl 5.30 was never really pushed to DockerHub, so "latest" should refer
to 5.26 instead


phracek:
Image is present here: https://hub.docker.com/r/centos/perl-530-centos7

PR #185, opened 2020-09-17
[DO NOT MERGE] Document usage with a Dockerfile

fila43:
This PR requires sclorg/container-common-scripts#161 to be merged before, because it uses a new function being added in that PR.

This adds the following:

*    Dockerfile examples for a variant that uses s2i scripts and a variant without s2i scripts
 *   Documentation enhancement that describes how the package can be used using a Dockerfile
  *  a test for the example Dockerfiles

This is a similar pull-request as [sclorg/httpd-container#99.](https://github.com/sclorg/httpd-container/pull/99) and [sclorg/nodejs-container#247](https://github.com/sclorg/s2i-nodejs-container/pull/247)

To see the formatted README.md, see https://github.com/fila43/s2i-perl-container/blob/master/5.30/README.md and it is also visible on the catalog, which will be the most important view: https://catalog.redhat.com/software/containers/detail/5d3fff015a13461f5fb8635a

[DO NOT MERGE] at this point: Since a similar change will be done to other images, let's make sure it is good enough -- we should get review from other developers/QE first, then by PM, and finally by a the Doc team.

### DIFF
--- a/imagestreams/perl-rhel-aarch64.json
+++ b/imagestreams/perl-rhel-aarch64.json
@@ -1,0 +1,53 @@
+{
+  "kind": "ImageStream",
+  "apiVersion": "v1",
+  "metadata": {
+    "name": "perl",
+    "annotations": {
+      "openshift.io/display-name": "Perl"
+    }
+  },
+  "spec": {
+    "tags": [
+      {
+        "name": "latest",
+        "annotations": {
+          "openshift.io/display-name": "Perl (Latest)",
+          "openshift.io/provider-display-name": "Red Hat, Inc.",
+          "description": "Build and run Perl applications on UBI. For more information about using this builder image, including OpenShift considerations, see https://github.com/sclorg/s2i-perl-container/blob/master/5.30/README.md.\n\nWARNING: By selecting this tag, your application will automatically update to use the latest version of Perl available on OpenShift, including major version updates.",
+          "iconClass": "icon-perl",
+          "tags": "builder,perl",
+          "supports":"perl",
+          "sampleRepo": "https://github.com/sclorg/dancer-ex.git"
+        },
+        "from": {
+          "kind": "ImageStreamTag",
+          "name": "5.26-ubi8"
+        },
+        "referencePolicy": {
+          "type": "Local"
+        }
+      },
+      {
+        "name": "5.26-ubi8",
+        "annotations": {
+          "openshift.io/display-name": "Perl 5.26 (UBI 8)",
+          "openshift.io/provider-display-name": "Red Hat, Inc.",
+          "description": "Build and run Perl 5.26 applications on UBI 8. For more information about using this builder image, including OpenShift considerations, see https://github.com/sclorg/s2i-perl-container/blob/master/5.26/README.md.",
+              "iconClass": "icon-perl",
+          "tags": "builder,perl",
+          "supports":"perl:5.26,perl",
+          "version": "5.26",
+          "sampleRepo": "https://github.com/sclorg/dancer-ex.git"
+        },
+        "from": {
+          "kind": "DockerImage",
+          "name": "registry.redhat.io/ubi8/perl-526:latest"
+        },
+        "referencePolicy": {
+          "type": "Local"
+        }
+      }
+    ]
+  }
+}


5.26-mod_fcgid/README.md
```diff
@@ -13,41 +13,131 @@ Note: while the examples in this README are calling `podman`, you can replace an
...
 the nodejs itself is included just to make the npm work.
 
-Usage
+Usage in Openshift
 ---------------------
-For this, we will assume that you are using the `rhscl/perl-526-rhel7 image`, available via `perl:5.26` imagestream tag in Openshift.
-Building a simple [perl-sample-app](https://github.com/sclorg/s2i-perl-container/tree/master/5.26/test/sample-test-app) application
-in Openshift can be achieved with the following step:
-
+ In this example, we will assume that you are using the `rhscl/perl-530-rhel7` image, available via `perl:5.26` imagestream tag in Openshift.
```
hhorak, line 31: A typo, `rhscl/perl-530-rhel7` instead of `rhscl/perl-526-rhel7`.

fila43:
[test]

5.26-mod_fcgid/README.md
```diff
@@ -13,41 +13,131 @@ Note: while the examples in this README are calling `podman`, you can replace an
...
+FROM centos/perl-526-centos7
+
+# Add application sources
+ADD app-src .
+
+USER 0
+
+# Install the dependencies
+RUN export PATH=${PATH}:/opt/rh/rh-perl526/root/usr/bin/&& \
+     export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/opt/rh/rh-perl526/root/usr/lib64 && \
```
ppisar, line 102: Setting the PATH and LD_LIRBARY_PATH is only necessaryfor RHSCL-based images (i.e. for RHEL-7).  But 5.26-mod_fcgid/README.md is a UBI8 image and it does not use RHSCL, hence setting the variables is not needed. Actually those paths do not exist in the image.

Basically I object against adding examples for 5.26/README.md into 5.26-mod_fcgid/README.md. It only confuses the user.

fila43:
[test]

examples/from-dockerfile/test
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+set -xe
+
+function ct_test_app_dockerfile() {
```
hhorak, line 5: This function can be now taken from common/ sub-module (once it is updated) and doesn't need to be defined here: https://github.com/sclorg/container-common-scripts/blob/master/test-lib.sh#L831
Just be careful that the latest variant of the function added one argument, that fills in the `$app_dir` variable from 4th function argument.

examples/from-dockerfile/test
```diff
@@ -0,0 +1,27 @@
...
+  sleep 2
+  ip=$(docker inspect --format='{{.NetworkSettings.IPAddress}}' "${app_cont_name}")
+  curl "http://$ip:${port}" | grep "${expected_text}"
+  docker kill "${app_cont_name}"
+  sleep 2
+  docker rmi "${app_image_name}"
+  rm -rf "${app_dir}"
+}
+
+ct_test_app_dockerfile Dockerfile 'https://github.com/sclorg/dancer-ex.git' 'Welcome to your Perl application on OpenShift' 8080
```
hhorak, line 26: These calls should be ideally moved to `test/run`, so it's all called during the PR testing, and the whole file `examples/from-dockerfile/test ` can be removed. I think the original reason for this file was to test these changes quickly, but the point is to make these tests be run during the general testing. I know it was done the same in other prepared examples, so sorry for confusing.

fila43:
[test]

fila43:
[test]

fila43:
[test]

fila43:
[test]

fila43:
[test]

fila43:
[test]

hhorak:
A weird tests behaviour can be caused by this:
https://github.com/sclorg/container-common-scripts/pull/173/files
(consider updating common sub-module)

fila43:
[test]

ppisar:
I'd like note that centos/perl-526-centos7 that you put everywhere is going to expire in May 2020. I'd recommend you centos/perl-530-centos7 that is supported until June 2024 <https://access.redhat.com/support/policy/updates/rhscl-rhel7>.

fila43:
[test]

fila43:
[test]

Issue #189, opened 2020-10-07
Test cases described as [PASSED] even though they failed


pkubatrh:
The code in master is now failing tests because of missing fix from common scripts. However, the failing test is still described as [PASSED]:

```
Version to check is 5.30.
Checking file imagestreams/perl-rhel.json.
Checking file imagestreams/perl-rhel-aarch64.json.
The latest version is not present in imagestreams/perl-rhel-aarch64.json or in latest tag.
Checking file imagestreams/perl-centos.json.
[PASSED] test_latest_imagestreams

Tests for centos/perl-526-centos7:5.26 failed.
make: *** [common/common.mk:94: test] Error 1
```

@phracek Can you take a look?

phracek:
I guess this can be closed.

PR #191, opened 2020-10-08
Fix check_result function

phracek:
Function check_result now returns return code so the test suite can catch failures.

### DIFF
--- a/5.26-mod_fcgid/test/run
+++ b/5.26-mod_fcgid/test/run
@@ -369,18 +369,15 @@ test_latest_imagestreams() {
 }
 
 function run_all_tests() {
-  local suite_result=0
   for test_case in $TEST_LIST; do
     : "Running test $test_case ...."
     if $test_case; then
       printf -v test_short_summary "${test_short_summary}[PASSED] $test_case\n"
     else
       printf -v test_short_summary "${test_short_summary}[FAILED] $test_case\n"
-      suite_result=1
-      [ -n "${FAIL_QUICKLY:-}" ] && echo "$sum" && return $suite_result
+      [ -n "${FAIL_QUICKLY:-}" ] && return 1
     fi
   done;
-  return $suite_result
 }
 
 # Run the chosen tests
--- a/5.26-mod_fcgid/test/run
+++ b/5.26-mod_fcgid/test/run
@@ -90,8 +90,7 @@ cleanup() {
 check_result() {
   local result="$1"
   if [[ "$result" != "0" ]]; then
-    TESTSUITE_RESULT=1
-    suite_result=1
+    TESTCASE_RESULT=1
   fi
 }
 
@@ -371,12 +370,13 @@ test_latest_imagestreams() {
 function run_all_tests() {
   for test_case in $TEST_LIST; do
     : "Running test $test_case ...."
-    suite_result=0
+    TESTCASE_RESULT=0
     $test_case
-    if [ $suite_result -eq 0 ]; then
+    if [ $TESTCASE_RESULT -eq 0 ]; then
       printf -v test_short_summary "${test_short_summary}[PASSED] $test_case\n"
     else
       printf -v test_short_summary "${test_short_summary}[FAILED] $test_case\n"
+      TESTSUITE_RESULT=1
       [ -n "${FAIL_QUICKLY:-}" ] && return 1
     fi
   done;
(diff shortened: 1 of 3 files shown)


phracek:
[test]

phracek:
The imagestream failure is fixed in container-common-scripts. https://github.com/sclorg/container-common-scripts/pull/175

pkubatrh:
This will be enough for cases that have `check_result` as the last call in their body. The return value from `check_result` will still be ignored for functions that have some cleanup actions at the end.

I see `check_result` sets `TESTSUITE_RESULT` which is then used for the final exit status of the test suite. `run_all_tests` has `suite_result` which is returned at its end, but not used at all.
How about combining these two variables? Use one to return the test case result to `run_all_tests` function, and then use its return value to set the scripts exit status. WDYT?

pkubatrh:
Fixes: https://github.com/sclorg/s2i-perl-container/issues/189

phracek:
> This will be enough for cases that have `check_result` as the last call in their body. The return value from `check_result` will still be ignored for functions that have some cleanup actions at the end.
> 
> I see `check_result` sets `TESTSUITE_RESULT` which is then used for the final exit status of the test suite. `run_all_tests` has `suite_result` which is returned at its end, but not used at all.
> How about combining these two variables? Use one to return the test case result to `run_all_tests` function, and then use its return value to set the scripts exit status. WDYT?

`suite_result` does not make sense at all, I will remove it.
I agree `check_result` will set `TESTSUITE_RESULT` and if [ -n FAIL_QUICK ] is set, then immediatelly return 1 and not `sum` and `suite_result`.

pkubatrh:
That does not fix the problem with return value from `check_result` being possibly ignored. The test case might still be described as [PASSED] even if it failed (`TESTSUITE_RESULT` is 1). For example, after applying the following:
```
diff --git a/5.30/test/run b/5.30/test/run
index f05aecc..8167be4 100755
--- a/5.30/test/run
+++ b/5.30/test/run
@@ -14,12 +14,6 @@ image_dir=$(readlink -zf ${test_dir}/..)
 test_short_summary=''
 TESTSUITE_RESULT=0
 TEST_LIST="\
-test_sample_test_app
-test_binpath
-test_psgi
-test_psgi_variables
-test_warningonstderr
-test_npm
 test_latest_imagestreams
 "
 
@@ -354,6 +348,7 @@ test_latest_imagestreams() {
   result=$?
   popd >/dev/null || return 1
   check_result $result
+  info "Test complete"
 }
 
 function run_all_tests() {
```
results in:
```
$ OS=rhel7 VERSION=5.30 test/run
test/run: line 12: warning: command substitution: ignored null byte in input
test/run: line 13: warning: command substitution: ignored null byte in input

[INFO] Testing the latest version in imagestreams...

Version to check is 5.30.
Checking file imagestreams/perl-rhel.json.
Checking file imagestreams/perl-rhel-aarch64.json.
The latest version is not present in imagestreams/perl-rhel-aarch64.json or in latest tag.
Checking file imagestreams/perl-centos.json.

[INFO] Test complete...

[PASSED] test_latest_imagestreams

Tests for centos/perl-530-centos-candidate failed.
```

phracek:
[test]

5.26-mod_fcgid/test/run
```diff
@@ -91,6 +91,7 @@ check_result() {
   local result="$1"
   if [[ "$result" != "0" ]]; then
     TESTSUITE_RESULT=1
+    suite_result=1
```
pkubatrh, line 94: I would rename it slightly, so its similar to `TESTSUITE_RESULT` - all caps, maybe name change to `TESTCASE_RESULT` for clear difference?

5.26-mod_fcgid/test/run
```diff
@@ -368,18 +369,17 @@ test_latest_imagestreams() {
...
   for test_case in $TEST_LIST; do
     : "Running test $test_case ...."
-    if $test_case; then
+    suite_result=0
+    $test_case
+    if [ $suite_result -eq 0 ]; then
       printf -v test_short_summary "${test_short_summary}[PASSED] $test_case\n"
     else
       printf -v test_short_summary "${test_short_summary}[FAILED] $test_case\n"
-      suite_result=1
```
pkubatrh, line 378: How about setting `TESTSUITE_RESULT` over here, instead of in `check_result`. Having two result variable in the same place might be a bit confusing.

pkubatrh (commented):
Some minor suggestions

5.26-mod_fcgid/test/run
phracek: Make sense.

phracek:
[test]

phracek:
Tests are green. Merging so nighly builds are fine.

PR #192, opened 2020-10-12
[DO NOT MERGE] Document usage with a Dockerfile 

fila43:
This adds the following:

    Dockerfile examples for a variant that uses s2i scripts and a variant without s2i scripts
    Documentation enhancement that describes how the package can be used using a Dockerfile
    a test for the example Dockerfiles

This is a similar pull-request as sclorg/httpd-container#99. and sclorg/nodejs-container#247

To see the formatted README.md, see https://github.com/fila43/s2i-perl-container/blob/master/5.30/README.md and it is also visible on the catalog, which will be the most important view: https://catalog.redhat.com/software/containers/detail/5d3fff015a13461f5fb8635a

[DO NOT MERGE] at this point: Since a similar change will be done to other images, let's make sure it is good enough -- we should get a review from other developers/QE first, then by PM, and finally by a Doc team.


fila43:
[test]

fila43:
[test]

5.26-mod_fcgid/README.md
```diff
@@ -1,5 +1,5 @@
-Perl 5.26 container image
-=========================
+Perl 5.30 container image
+=================
 
 This container image includes Perl 5.26 as a [S2I](https://github.com/openshift/source-to-image) base image for your Perl 5.26 applications.
 Users can choose between RHEL, CentOS and Fedora based builder images.
```
ppisar: I don't what your goal is, but writing "Perl 5.30" into README for 5.26 is utterly wrong.

fila43: thanks, my mistake it was a typo

5.26-mod_fcgid/README.md
```diff
@@ -13,41 +13,131 @@ Note: while the examples in this README are calling `podman`, you can replace an
...
     ```
 
-The same application can also be built using the standalone [S2I](https://github.com/openshift/source-to-image) application on systems that have it available:
+**To access the application:**
+```
+$ oc get pods
+$ oc exec <pod> -- curl 127.0.0.1:8080
+```
+
+Source-to-Inage framework and scripts
```
ppisar, line 43: A typo. Should be Source-to-Image. On multiple places.

PR #194, opened 2020-10-20
Fix "own setup" example for mod_fcgid images

ppisar:
The examples introduced in commit
ad12949496e45c1a6f8fb9a48e0279ddf75c9299 presume mod_perl, but they do
not work with mod_fcgid.

This patch reworks the example to properly configure FCGI server from
a PSGI application where needed, i.e. in 5.26-mod_fcgid.


ppisar:
[test]

Issue #195, opened 2020-11-06
5.30-mod_fcgid is absent


yselkowitz:
Despite the release of RHEL/UBI 8.3 with a perl-530 mod_fcgid image, a corresponding directory is missing from this repository.

ppisar:
I will send a pull request.

ppisar:
@pkubatrh, what base image should I refer in FROM? ubi8/s2i-base:1 or ubi8/s2i-base:rhel8.3? The underlying perl:5.30 module is available since RHEL-8.3.

pkubatrh:
`ubi8/s2i-base:1`, the `:rhel8.3` tag is only available internally.

And thanks for taking care of the PR!

ppisar:
Fixed with 770674e7367ec9fd8e1b0e1553c98299cc90808d.

PR #197, opened 2020-11-09
Add Perl 5.30 with mod_fcgid for RHEL-8

ppisar:
5.30-mod_fcgid is built on Perl 5.30 and mod_fcgid from modules.

RHEL7 does not contain perl-FCGI. Thus 5.30-mod_fcgi is not supported
on RHEL7.

F33 and older Fedoras do not contain modular perl-FCGI. F34 does not
have s2i-base image. Thus 5.30-mod_fcgi is not supported on Fedora.

https://github.com/sclorg/s2i-perl-container/issues/195


ppisar:
The image probably won't build because a UBI8 repository <https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi8/8/x86_64/appstream/os> is missing vital modules. E.g. perl-IO-Socket-SSL:2.066.

pkubatrh:
Could we do the "git cp/mv dance" described at https://github.com/sclorg/welcome/blob/master/contribution.md#adding-a-new-version ? So we always have the history in the last version.

>The image probably won't build because a UBI8 repository is missing vital modules.

Would you have the full list of rpms that are missing from ubi8? I can follow up on getting them added to the repository so the image builds

ppisar:
The missing ones are perl-IO-Socket-SSL:2.066 and perl-libwww-perl:6.34
modularity modules. I've already reported the issue against a configuration of
the UBI composes.

ppisar:
How can I disable checks for rhel7? This docker file won't work on RHEL 7. The commit contains .exclude-rhel7 file, but it seems ignored. Can I merge the commit in spite of the failing test?

ppisar:
I made the git move dance. Although I think it's bad. It clutters the commit history and breaks a bisect.

ppisar:
The tests failed for perl-526 that I did not touch at all. And they failed only on one of the RHEL versions. Last time the same code failed on the other RHEL version. The CI infrastructure is unreliable.

PR #198, opened 2020-11-18
enable scl variables

fila43:
enable scl env variable, based on that is not necessary to use versioned from-dockerfiles examples

### DIFF
--- a/5.26/Dockerfile
+++ b/5.26/Dockerfile
@@ -7,7 +7,8 @@ EXPOSE 8080
 # Image metadata
 ENV PERL_VERSION=5.26 \
     PERL_SHORT_VER=526 \
-    NAME=perl
+    NAME=perl \
+    HTTPD_24=httpd24
 
 ENV SUMMARY="Platform for building and running Perl $PERL_VERSION applications" \
     DESCRIPTION="Perl $PERL_VERSION available as container is a base platform for \
@@ -16,7 +17,15 @@ Perl is a high-level programming language with roots in C, sed, awk and shell sc
 Perl is good at handling processes and files, and is especially good at handling text. \
 Perl's hallmarks are practicality and efficiency. While it is used to do a lot of \
 different things, Perl's most common applications are system administration utilities \
-and web programming."
+and web programming." \
+     PKGG_CONFIG_PATH=/opt/rh/$HTTPD_24/root/usr/lib64/pkgconfig \
+     PYTHONPATH=/opt/rh/$NODEJS_SCL/root/usr/lib/python2.7/site-packages \
+     PATH=/opt/rh/$NODEJS_SCL/root/usr/bin:/opt/app-root/src/bin:/opt/app-root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/local/bin:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/bin:/opt/rh/$HTTPD_24/root/usr/bin:/opt/rh/$HTTPD_24/root/usr/sbin:/opt/app-root/src/extlib/bin \
+     LD_LIBRARY_PATH=/opt/rh/$NODEJS_SCL/root/usr/lib64:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/lib64:/opt/rh/$HTTPD_24/root/usr/lib64 \
+     X_SCLS="rh-perl$PERL_SHORT_VER $NODEJS_SCL $HTTPD_24" \
+     LIBRARY_PATH=/opt/rh/$HTTPD_24/root/usr/lib64 \
+     MANPATH=/opt/rh/$NODEJS_SCL/root/usr/share/man:/opt/rh/$HTTPD_24/root/usr/share/man:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/share/man \
+     PERL5LIB=/opt/app-root/src/extlib/lib/perl5
 
 LABEL summary="$SUMMARY" \
       description="$DESCRIPTION" \
@@ -49,16 +58,15 @@ COPY ./root/ /
 # writeable as OpenShift default security model is to run the container under
 # random UID.
 RUN mkdir -p ${APP_ROOT}/etc/httpd.d && \
-    sed -i -f ${APP_ROOT}/etc/httpdconf.sed /opt/rh/httpd24/root/etc/httpd/conf/httpd.conf  && \
-    chmod -R og+rwx /opt/rh/httpd24/root/var/run/httpd ${APP_ROOT}/etc/httpd.d && \
+    sed -i -f ${APP_ROOT}/etc/httpdconf.sed /opt/rh/$HTTPD_24/root/etc/httpd/conf/httpd.conf  && \
+    chmod -R og+rwx /opt/rh/$HTTPD_24/root/var/run/httpd ${APP_ROOT}/etc/httpd.d && \
     chown -R 1001:0 ${APP_ROOT} && chmod -R ug+rwx ${APP_ROOT} && \
     rpm-file-permissions
 
 USER 1001
 
 # Enable the SCL for all bash scripts.
-ENV BASH_ENV=${APP_ROOT}/etc/scl_enable \
-    ENV=${APP_ROOT}/etc/scl_enable \
+ENV ENV=${APP_ROOT}/etc/scl_enable \
     PROMPT_COMMAND=". ${APP_ROOT}/etc/scl_enable"
 
 # Set the default CMD to print the usage of the language image
--- a/5.26/Dockerfile.rhel7
+++ b/5.26/Dockerfile.rhel7
@@ -7,7 +7,8 @@ EXPOSE 8080
 # Image metadata
 ENV PERL_VERSION=5.26 \
     PERL_SHORT_VER=526 \
-    NAME=perl
+    NAME=perl \
+    HTTPD_24=httpd24
 
 ENV SUMMARY="Platform for building and running Perl $PERL_VERSION applications" \
     DESCRIPTION="Perl $PERL_VERSION available as container is a base platform for \
@@ -16,7 +17,15 @@ Perl is a high-level programming language with roots in C, sed, awk and shell sc
 Perl is good at handling processes and files, and is especially good at handling text. \
 Perl's hallmarks are practicality and efficiency. While it is used to do a lot of \
 different things, Perl's most common applications are system administration utilities \
-and web programming."
+and web programming."\
+     PKGG_CONFIG_PATH=/opt/rh/$HTTPD_24/root/usr/lib64/pkgconfig \
+     PYTHONPATH=/opt/rh/$NODEJS_SCL/root/usr/lib/python2.7/site-packages \
+     PATH=/opt/rh/$NODEJS_SCL/root/usr/bin:/opt/app-root/src/bin:/opt/app-root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/local/bin:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/bin:/opt/rh/$HTTPD_24/root/usr/bin:/opt/rh/$HTTPD_24/root/usr/sbin:/opt/app-root/src/extlib/bin \
+     LD_LIBRARY_PATH=/opt/rh/$NODEJS_SCL/root/usr/lib64:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/lib64:/opt/rh/$HTTPD_24/root/usr/lib64 \
+     X_SCLS="rh-perl$PERL_SHORT_VER $NODEJS_SCL $HTTPD_24" \
+     LIBRARY_PATH=/opt/rh/$HTTPD_24/root/usr/lib64 \
+     MANPATH=/opt/rh/$NODEJS_SCL/root/usr/share/man:/opt/rh/$HTTPD_24/root/usr/share/man:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/share/man \
+     PERL5LIB=/opt/app-root/src/extlib/lib/perl5
 
 LABEL summary="$SUMMARY" \
       description="$DESCRIPTION" \
@@ -51,16 +60,15 @@ COPY ./root/ /
 # writeable as OpenShift default security model is to run the container under
 # random UID.
 RUN mkdir -p ${APP_ROOT}/etc/httpd.d && \
-    sed -i -f ${APP_ROOT}/etc/httpdconf.sed /opt/rh/httpd24/root/etc/httpd/conf/httpd.conf  && \
-    chmod -R og+rwx /opt/rh/httpd24/root/var/run/httpd ${APP_ROOT}/etc/httpd.d && \
+    sed -i -f ${APP_ROOT}/etc/httpdconf.sed /opt/rh/$HTTPD_24/root/etc/httpd/conf/httpd.conf  && \
+    chmod -R og+rwx /opt/rh/$HTTPD_24/root/var/run/httpd ${APP_ROOT}/etc/httpd.d && \
     chown -R 1001:0 ${APP_ROOT} && chmod -R ug+rwx ${APP_ROOT} && \
     rpm-file-permissions
 
 USER 1001
 
 # Enable the SCL for all bash scripts.
-ENV BASH_ENV=${APP_ROOT}/etc/scl_enable \
-    ENV=${APP_ROOT}/etc/scl_enable \
+ENV ENV=${APP_ROOT}/etc/scl_enable \
     PROMPT_COMMAND=". ${APP_ROOT}/etc/scl_enable"
 
 # Set the default CMD to print the usage of the language image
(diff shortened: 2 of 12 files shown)


fila43:
[test]

fila43:
[test]

ppisar:
If you preset SCL environment variables on Dockerfile level as if "scl enable" was invoked, all invocations of "scl enable" could be removed. It would be great to delete:

* root/opt/app-root/etc/scl_enable files
* "ENV BASH_ENV" line from the Dockerfiles

pkubatrh:
> If you preset SCL environment variables on Dockerfile level as if "scl enable" was invoked, all invocations of "scl enable" could be removed. It would be great to delete:

True, though the tests for this feature set the use case when `bash` is called inside the image as a correct (thanks to `scl_enable`) baseline to check against, so they would need to be rewritten to be a bit smarter (by providing list of expected SCLs enabled provided in the image's test suite for example) if we wanted to remove the `scl_enable` files entirely.
I wonder if the `scl_enable` files might possibly used by some of our advanced users. I do not think we document it anywhere so I find it unlikely, but would still see removing the files as a bit of a risk because of this as well.

fila43:
[test]

fila43:
It sounds fine, but do you see removing this stuff necessary for now?  From my point of view, the worst case is enabling scl twice. And only centos7 and rhel7 are affected. I think it should be removed later, together with all other containers.

ppisar:
My point is that you make it more complicated and that keeping the scl_enable call masks all mistakes you can do in the Dockerfiles.

5.26/Dockerfile
```diff
@@ -16,7 +16,14 @@ Perl is a high-level programming language with roots in C, sed, awk and shell sc
 Perl is good at handling processes and files, and is especially good at handling text. \
 Perl's hallmarks are practicality and efficiency. While it is used to do a lot of \
 different things, Perl's most common applications are system administration utilities \
-and web programming."
+and web programming." \
+     PKG_CONFIG_PATH=/opt/rh/httpd24/root/usr/lib64/pkgconfig \
+     PYTHONPATH=/opt/rh/$NODEJS_SCL/root/usr/lib/python2.7/site-packages \
+     PATH=$PATH:/opt/rh/$NODEJS_SCL/root/usr/bin:/opt/rh/httpd24/root/usr/bin:/opt/rh/httpd24/root/usr/sbin:/opt/app-root/src/bin:/opt/app-root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/rh/$SCL_NODEJS/root/usr/bin:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/local/bin:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/bin:/opt/rh/httpd24/root/usr/bin:/opt/rh/httpd24/root/usr/sbin:/opt/rh-perl526/root/bin \
```
ppisar, line 22: Duplicated values:

/opt/rh/httpd24/root/usr/bin
/opt/rh/httpd24/root/usr/sbin

Is /opt/rh/$NODEJS_SCL/root/usr/bin and /opt/rh/$SCL_NODEJS/root/usr/bin diffrent?

/opt/rh-perl526/root/bin should be spelled with $PERL_SHORT_VER.

A path for user-installed executables, /opt/app-root/src/extlib/bin, is missing.

5.26/Dockerfile
```diff
@@ -16,7 +16,14 @@ Perl is a high-level programming language with roots in C, sed, awk and shell sc
...
+     LD_LIBRARY_PATH=/opt/rh/$NODEJS_SCL/root/usr/lib64:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/lib64:/opt/rh/httpd24/root/usr/lib64 \
+     X_SCLS=rh-perl$PERL_SHORT_VER \
```
ppisar, line 24: httpd24 and $NODEJS_SCL is missing from X_SCLS.

5.26/Dockerfile
```diff
@@ -16,7 +16,14 @@ Perl is a high-level programming language with roots in C, sed, awk and shell sc
...
+     LIBRARY_PATH=/opt/rh/httpd24/root/usr/lib64 \
```
ppisar, line 25: rh-perl$PERL_SHORT_VER and $NODEJS is misssing from LIBRARY_PATH.

5.26/Dockerfile
```diff
@@ -16,7 +16,14 @@ Perl is a high-level programming language with roots in C, sed, awk and shell sc
...
+     MANPATH=/opt/rh/$NODEJS_SCL/root/usr/share/man:/opt/rh/httpd24/root/usr/share/man
```
ppisar, line 26: rh-perl$PERL_SHORT_VER is missing from MANPATH

5.26/Dockerfile
```diff
@@ -16,7 +16,14 @@ Perl is a high-level programming language with roots in C, sed, awk and shell sc
...
```
ppisar, line 27: PERL5LIB=/opt/app-root/src/extlib/lib/perl5 for user-installed Perl modules is missing.

5.26/Dockerfile.rhel7
```diff
@@ -16,7 +16,14 @@ Perl is a high-level programming language with roots in C, sed, awk and shell sc
...
 different things, Perl's most common applications are system administration utilities \
-and web programming."
+and web programming."\
+     PKG_CONFIG_PATH=/opt/rh/httpd24/root/usr/lib64/pkgconfig \
+     PYTHONPATH=/opt/rh/$NODEJS_SCL/root/usr/lib/python2.7/site-packages \
+     PATH=$PATH:/opt/rh/$NODEJS_SCL/root/usr/bin:/opt/rh/httpd24/root/usr/bin:/opt/rh/httpd24/root/usr/sbin:/opt/app-root/src/bin:/opt/app-root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/rh/$SCL_NODEJS/root/usr/bin:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/local/bin:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/bin:/opt/rh/httpd24/root/usr/bin:/opt/rh/httpd24/root/usr/sbin:/opt/rh-perl526/root/bin \
+     LD_LIBRARY_PATH=/opt/rh/$NODEJS_SCL/root/usr/lib64:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/lib64:/opt/rh/httpd24/root/usr/lib64 \
+     X_SCLS=rh-perl$PERL_SHORT_VER \
+     LIBRARY_PATH=/opt/rh/httpd24/root/usr/lib64 \
+     MANPATH=/opt/rh/$NODEJS_SCL/root/usr/share/man:/opt/rh/httpd24/root/usr/share/man
```
ppisar, line 26: The same issues as in 5.26/Dockerfile.

5.30/Dockerfile
```diff
@@ -16,7 +16,14 @@ Perl is a high-level programming language with roots in C, sed, awk and shell sc
...
 different things, Perl's most common applications are system administration utilities \
-and web programming."
+and web programming." \
+     PKG_CONFIG_PATH=/opt/rh/httpd24/root/usr/lib64/pkgconfig \
+     PYTHONPATH=/opt/rh/$NODEJS_SCL/root/usr/lib/python2.7/site-packages \
+     PATH=$PATH:/opt/rh/$NODEJS_SCL/root/usr/bin:/opt/rh/httpd24/root/usr/bin:/opt/rh/httpd24/root/usr/sbin:/opt/app-root/src/bin:/opt/app-root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/rh/$SCL_NODEJS/root/usr/bin:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/local/bin:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/bin:/opt/rh/httpd24/root/usr/bin:/opt/rh/httpd24/root/usr/sbin:/opt/rh-perl526/root/bin \
+     LD_LIBRARY_PATH=/opt/rh/$NODEJS_SCL/root/usr/lib64:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/lib64:/opt/rh/httpd24/root/usr/lib64 \
+     X_SCLS=rh-perl$PERL_SHORT_VER \
+     LIBRARY_PATH=/opt/rh/httpd24/root/usr/lib64 \
+     MANPATH=/opt/rh/$NODEJS_SCL/root/usr/share/man:/opt/rh/httpd24/root/usr/share/man
```
ppisar, line 26: The same issues as in 5.26/Dockerfile.

5.30/Dockerfile.rhel7
```diff
@@ -16,7 +16,14 @@ Perl is a high-level programming language with roots in C, sed, awk and shell sc
...
 different things, Perl's most common applications are system administration utilities \
-and web programming."
+and web programming." \
+     PKG_CONFIG_PATH=/opt/rh/httpd24/root/usr/lib64/pkgconfig \
+     PYTHONPATH=/opt/rh/$NODEJS_SCL/root/usr/lib/python2.7/site-packages \
+     PATH=$PATH:/opt/rh/$NODEJS_SCL/root/usr/bin:/opt/rh/httpd24/root/usr/bin:/opt/rh/httpd24/root/usr/sbin:/opt/app-root/src/bin:/opt/app-root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/rh/$SCL_NODEJS/root/usr/bin:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/local/bin:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/bin:/opt/rh/httpd24/root/usr/bin:/opt/rh/httpd24/root/usr/sbin:/opt/rh-perl526/root/bin \
+     LD_LIBRARY_PATH=/opt/rh/$NODEJS_SCL/root/usr/lib64:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/lib64:/opt/rh/httpd24/root/usr/lib64 \
+     X_SCLS=rh-perl$PERL_SHORT_VER \
+     LIBRARY_PATH=/opt/rh/httpd24/root/usr/lib64 \
+     MANPATH=/opt/rh/$NODEJS_SCL/root/usr/share/man:/opt/rh/httpd24/root/usr/share/man
```
ppisar, line 26: The same issues as in 5.26/Dockerfile.

5.26/Dockerfile
fila43: It could be added, but actually scl_enable doesn't enable this path

5.26/Dockerfile
fila43: Is it necessary? Because scl_enable doesn't add rh-perl$PERL_SHORT_VER and $NODEJS into LIBRARY_PATH

ppisar: Of course "scl enable rh-perl526" amends LD_LIBRARY_PATH:

~~~~
# echo $LD_LIBRARY_PATH

# scl enable rh-perl526 'echo $LD_LIBRARY_PATH'
/opt/rh/rh-perl526/root/usr/lib64
~~~~

5.26/Dockerfile
ppisar: You are right that scl_enable does not do that. It is done in s2i/bin/run script. But I thought that a purpose of your change is have the same environment in both run and exec container modes. Without the paths Perl modules and scripts installed at the time of building an application image won't be available in the exec mode.

5.26/Dockerfile
fila43: I agree  LD_LIBRARY_PATH contains rh-perl526, but LIBRARY_PATH doesn't.

5.26/Dockerfile
fila43: ok thanks, that make sense

fila43:
[test]

5.26/Dockerfile
```diff
@@ -57,7 +65,7 @@ RUN mkdir -p ${APP_ROOT}/etc/httpd.d && \
 USER 1001
 
 # Enable the SCL for all bash scripts.
-ENV BASH_ENV=${APP_ROOT}/etc/scl_enable \
+ENV BASH_ENV \
```
phracek, line 68: BASH_ENV is a dummy variable? Can it be removed? I have lost the content here.

5.26/Dockerfile
```diff
@@ -16,7 +16,15 @@ Perl is a high-level programming language with roots in C, sed, awk and shell sc
 Perl is good at handling processes and files, and is especially good at handling text. \
 Perl's hallmarks are practicality and efficiency. While it is used to do a lot of \
 different things, Perl's most common applications are system administration utilities \
-and web programming."
+and web programming." \
+     PKGG_CONFIG_PATH=/opt/rh/httpd24/root/usr/lib64/pkgconfig \
+     PYTHONPATH=/opt/rh/$NODEJS_SCL/root/usr/lib/python2.7/site-packages \
+     PATH=/opt/rh/$NODEJS_SCL/root/usr/bin:/opt/app-root/src/bin:/opt/app-root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/local/bin:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/bin:/opt/rh/httpd24/root/usr/bin:/opt/rh/httpd24/root/usr/sbin \
+     LD_LIBRARY_PATH=/opt/rh/$NODEJS_SCL/root/usr/lib64:/opt/rh/rh-perl$PERL_SHORT_VER/root/usr/lib64:/opt/rh/httpd24/root/usr/lib64 \
```
phracek, line 23: httpd24 is defined several times. What about defining a variable like HTTPD_24=httpd24 and then use only $HTTPD_24?

5.26/Dockerfile.rhel7
```diff
@@ -16,7 +16,15 @@ Perl is a high-level programming language with roots in C, sed, awk and shell sc
 Perl is good at handling processes and files, and is especially good at handling text. \
 Perl's hallmarks are practicality and efficiency. While it is used to do a lot of \
 different things, Perl's most common applications are system administration utilities \
-and web programming."
+and web programming."\
+     PKGG_CONFIG_PATH=/opt/rh/httpd24/root/usr/lib64/pkgconfig \
```
phracek, line 20: The same here with httpd24 like before.

test/run
```diff
@@ -344,6 +345,22 @@ test_npm() {
...
+    ct_enable_cleanup
+
+    info "Testing variable presence during \`docker exec\`"
+    ct_check_exec_env_vars
+    check_result $?
+
+    info "Checking if all scl variables are defined in Dockerfile"
+    ct_check_scl_enable_vars
+    check_result $?
+ fi
```
phracek, line 361: `space` is missing. Just bash style.

phracek (changes requested):

phracek:
Thanks for driving this PR!. Cool.

test/run
fila43: Thanks for the notes!

fila43:
[test]

fila43:
@ppisar now it should be ready, I have disabled scl_enable, so paths shouldn't be duplicated and I have also enabled PERL5LIB, current env output should look the same as after the application of scl_enable.

ppisar:
I can see you set PERL5LIB=/opt/app-root/src/extlib/lib/perl5, but you don't
add /opt/app-root/src/extlib/bin into PATH. Please add it to PATH and remove
the the two settings at the beggining of s2i/bin/run.

fila43:
[test]

ppisar:
I can see that the tests passed in RHEL and CenOS. Fedora does not have
software collection, so the failure is expected. If you are ready, you can
merge it.

fila43:
ok, great, I looked at Fedora and it seems that adding PERL5LIB=/opt/app-root/src/extlib/lib/perl5 and /opt/app-root/src/extlib/bin into PATH  helps and tests pass, what is your opinion in case of adding this stuff into Fedora's dockerfile.

ppisar:
There is no perl 5.26 in Fedora. There is perl 5.30 in Fedora 32, but it isn't
in Fedora 33 which is used in the Dockerfiles now. Thus having any Dockerfiles
for Fedora does not make sense. If you built such a Perl image, it would be
a lie because a name of image would not match perl inside.

phracek:
> There is no perl 5.26 in Fedora. There is perl 5.30 in Fedora 32, but it isn't in Fedora 33 which is used in the Dockerfiles now. Thus having any Dockerfiles for Fedora does not make sense. If you built such a Perl image, it would be a lie because a name of image would not match perl inside.

As far as I know, F31 and F32 contain perl-30. We can change Dockerfile into f32/s2i-base. But this does not bring any new things.
It would be great to have Fedora tests passing as well. :)

phracek:
> I can see that the tests passed in RHEL and CenOS. Fedora does not have software collection, so the failure is expected. If you are ready, you can merge it.

You are right, Fedora does not have softw are collections, but it uses standard perl distributions, e.g. 5-30 in F31 and F32. 
Dockerfile.fedora is not changed, thus Failure is not expected.

ppisar:
I see. Then add setting the variables in Dockerfile.fedora.

fila43:
[test]

ppisar:
Great. Can I merge it?

fila43:
Yes, now it's ready, thanks for your cooperation!

PR #199, opened 2020-12-01
Use quay io instead docker

phracek:
This commit adds support for Quay.io registry instead of Docker.io registry.
Docker.io registry adds rate limits which sometimes blocks our test suite.


phracek:
[test]

phracek:
RHEL8 failure is expected because of `cidfile` issue.

fila43:
looks fine

ppisar:
No issue from Perl point of view.

PR #201, opened 2020-12-16
Use Quay.io in imagestreams

phracek:


ppisar:
No issue from Perl point of view.

PR #203, opened 2021-02-04
Fix testcase calling

phracek:
Add check_result right after testcase call.

This commit fixes also the case in case of
testcase is called but does not exist


phracek:
[test]

phracek:
This PR is divided into two commits.
First, commit deduplicates run and run-modfcid scripts.

The second commit does a few changes in order to fix proper calling testcase functions.

@fila43 @pkubatrh Please take a look at it. Thanks

phracek:
[test]

phracek:
[test]

ppisar:
So you moved an expected test output (./5.30/test/run) to a common ./test place, but you kept the test input (./5.30/test/*) in every perl version? I don't like the split. Could you please keep them together?

phracek:
[test]

ppisar:
That looks good. Thanks. You can merge it if you are ready.

PR #205, opened 2021-02-22
Deprecate perl-5.26 imagestreams for RHEL7 and CentOS7

phracek:

### DIFF
--- a/imagestreams/perl-centos.json
+++ b/imagestreams/perl-centos.json
@@ -107,46 +107,6 @@
         "referencePolicy": {
           "type": "Local"
         }
-      },
-      {
-        "name": "5.26-el7",
-        "annotations": {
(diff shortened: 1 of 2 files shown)


phracek:
[test][test-openshift]

ppisar:
rh-perl526 software collection for RHEL 7 is supported up to May 2021
<https://access.redhat.com/support/policy/updates/rhscl-rhel7>. Isn't
too early for removing it from OpenShift templates? Otherwise I have no
problems with this patch from Perl point of view.

pkubatrh:
@ppisar  The reason we are doing it early is so that it does not end up in the next release of Openshift after the images are EOL (the imagestream sync will be done some time before GA).

pkubatrh:
[test-openshift]

phracek:
Let's one more try.

[test-openshift]

PR #213, opened 2021-09-21
Fix README files

michal-josef-spacek:
There were information about nodejs, probably copy/paste from other
repository.


phracek (approved):
LGTM.

pkubatrh:
Note that the part about npm was there on purpose, as the perl images are built on top of s2i-base images which install npm for use in other layers. Not sure if this is important for the perl image to document.

https://github.com/sclorg/s2i-perl-container/commit/f7a2af17a7642441fb50699bb352f832b41a387a#diff-45fe7cda05c1b2c01eb3595020c734e7b84b9c1ce2f8b31aa1e82d7fe2e90446

PR #215, opened 2021-10-25
Fix example of build in documentation

michal-josef-spacek:
There are some issues:
* multiple use of CMD not allowed
* sed -i '1i...' don't work, you need to describe -e before command
* sed on non-exiting file isn't possible

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        version: [5.30]
+        version: ["5.30"]
     steps:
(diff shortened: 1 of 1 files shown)


michal-josef-spacek:
This fix is for `podman`, which is a bit different than `docker`. Still, documentation for docker is same, one command per Dockerfile file.

There is missing change in test because there is test 'test_from_dockerfile' which contain same command. I am preparing change.

michal-josef-spacek:
Twice (and more) usage of CMD is silently not working in `docker`. Sic!

michal-josef-spacek:
New PR https://github.com/sclorg/s2i-perl-container/pull/217

PR #216, opened 2021-10-25
Fix two bash warnings

michal-josef-spacek:
When calling readline do not use the -z option.
From the readline manpage
           -z, --zero
                  end each output line with NUL, not newline

This fixes the following two bash warnings

test/run: line 11: warning: command substitution: ignored null byte in
input
test/run: line 12: warning: command substitution: ignored null byte in
input


phracek:
[test]

zmiklank (approved):
Thanks for PR. LGTM.

PR #217, opened 2021-11-09
Fix example of build in documentation and code

michal-josef-spacek:
There are some issues:
* multiple use of CMD not allowed
* sed -i '1i...' don't work, you need to describe -e before command
* sed on non-exiting file isn't possible


michal-josef-spacek:
[test]

phracek:
[test]

phracek (approved):
Tests are passing.

LGTM

PR #219, opened 2021-11-09
Remove trailing `docker rm`

michal-josef-spacek:
If we have `docker run --rm ...` we don't need to use `docker rm` in
cleanup.


phracek:
[test]

michal-josef-spacek:
This commit fixes issue in log from Testing farm:
```
05f91a02f9056407f1f94ec1edcdb455073e0e266ca6b55dcaffcf8c6f5f0f7f
Error response from daemon: No such container: 05f91a02f9056407f1f94ec1edcdb455073e0e266ca6b55dcaffcf8c6f5f0f7f
```

michal-josef-spacek:
@phracek Could we merge this fix? It is very unpleasant to develop with this kind of issues.

michal-josef-spacek:
@zmiklank 
in this case there were bad situation:
```
docker run --user=100001 --rm --cidfile=${cid_file} ${IMAGE_NAME}-testapp
...
docker stop # Removes container because --rm in previous - You could check by docker container ls -a
docker rm $(cat $cid_file) # Removes again -> warning
```
In case of `common/test-lib.sh` in the `ct_cleanup` function)
If there is command with --rm before them, then yes. btw: Command `docker inspect -f '{{.State.ExitCode}}' "$container"` will fail, because container doesn't exists.

ad test_docker_run_usage())
Good catch. There is missing --rm in `docker run` if we want to go this way.

I believe that we need to go in way of `common/` functions.

zmiklank:
@michal-josef-spacek 

There is only one call of `docker run` without the `--rm` option in the `common/` sources, but it's there for a reason, so I think, that following the `common/` way is correct. 

However I propose to add the --rm option to the `test_docker_run_usage()`, as cleaning up for this function is not in any way done with `test-lib.sh`'s `ct_cleanup()`.

pkubatrh:
Since we are only trying to remove a container after we check for its existence, should we not rather figure out why the remove is called for containers that are already removed?

michal-josef-spacek:
@zmiklank @pkubatrh I think, that best way is improve test-lib.sh `ct_cleanup()` to accept cleanup process in situation when container is removed. So this mean that `docker inspect` is failing. And rewrite this scripts to `ct_cleanup()`.
Is this acceptable?

zmiklank:
@michal-josef-spacek I like the idea of `run` scripts to use `test-lib.sh`'s cleaning functions.
However, if I understand correctly, the `docker inspect` shouldn't end with fail in this case, but the following line would fail, if already removed container is again removed when stopped:
```
  42     docker rm -v "$container"
```

@pkubat Do you mean that you would want to unify the container removals in the `test-lib.sh`? (Either always use --rm, and do not remove it in ct_cleanup, or do not use --rm at all, and remove all containers in the end in ct_cleanup)

pkubatrh:
>I think, that best way is improve test-lib.sh ct_cleanup() to accept cleanup process in situation when container is removed. So this mean that docker inspect is failing. And rewrite this scripts to ct_cleanup().

Sure, this seems like a good idea to me as well. Let us make sure `ct_cleanup` is able to ignore already removed containers and then replace perl container's clean up action with the library one.

michal-josef-spacek:
@zmiklank 
Example:
```
> cat Dockerfile
FROM alpine:latest

CMD echo "1"
```
```
skim@woman:/home/skim/lang/docker/check# docker build -t foo .
Sending build context to Docker daemon  3.072kB
Step 1/2 : FROM alpine:latest
 ---> c059bfaa849c
Step 2/2 : CMD echo "1"
 ---> Using cache
 ---> e575b3aa212b
Successfully built e575b3aa212b
Successfully tagged foo:latest
skim@woman:/home/skim/lang/docker/check# docker run foo
1
skim@woman:/home/skim/lang/docker/check# docker container ls
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES
skim@woman:/home/skim/lang/docker/check# docker container ls -a
CONTAINER ID   IMAGE     COMMAND                   CREATED         STATUS                     PORTS     NAMES
bb30ceec7ad1   foo       "/bin/sh -c 'echo \"1…"   5 seconds ago   Exited (0) 4 seconds ago             pedantic_visvesvaraya
skim@woman:/home/skim/lang/docker/check# docker container rm bb30ceec7ad1
bb30ceec7ad1
skim@woman:/home/skim/lang/docker/check# docker container ls -a
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES
```
With --rm
```
skim@woman:/home/skim/lang/docker/check# docker run --rm foo
1
skim@woman:/home/skim/lang/docker/check# docker container ls
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES
skim@woman:/home/skim/lang/docker/check# docker container ls -a
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES
```

This mean, that `docker inspect` for first is working, for second not (when container id have from cid_file).

zmiklank:
@michal-josef-spacek 

Oh, I see now, thanks for the explanation!

michal-josef-spacek:
PR for fix in common: https://github.com/sclorg/container-common-scripts/pull/241

phracek:
[test-all]

michal-josef-spacek:
The issue is still here, only in test/run-modfcgid. I think, that we need to rewrite test/run-modfcgid now to the same style as test/run now.

phracek:
@michal-josef-spacek Can you please rebase it against master?

phracek:
> The issue is still here, only in test/run-modfcgid. I think, that we need to rewrite test/run-modfcgid now to the same style as test/run now.

I would create a new pull request for rewriting `test/run-modfcgid`.

michal-josef-spacek:
@phracek done

zmiklank (approved):

phracek:
[test-all]

phracek (approved):

PR #221, opened 2021-11-16
Support PSGI reload

michal-josef-spacek:
This is functionality like PERL_APACHE2_RELOAD in containers with mod_perl and raw calling of *.pl applications


michal-josef-spacek:
I am working on test for this functionality

phracek:
[test]

PR #229, opened 2022-03-21
use composite build-and-push github action

zmiklank:
and rebuild image every wednesday


phracek (approved):
LGTM. Thanks for the PR!

PR #234, opened 2022-05-16
Support perl 5.32 and Fedora

phracek:
This pull request adds support for Fedora Perl-5.32 container image.

Signed-off-by: Petr "Stone" Hracek <phracek@redhat.com>

PR #239, opened 2022-05-27
Fix PATH variable

michal-josef-spacek:
$PATH must be on the end


michal-josef-spacek:
[test]

phracek:
This pull request is already fixed by https://github.com/sclorg/s2i-perl-container/pull/234

PR #245, opened 2022-06-06
run all openshift4 tests even when one fails

zmiklank:


zmiklank:
[test-all]

test/run-openshift-remote-cluster
```diff
@@ -13,7 +13,14 @@ source "${THISDIR}/test-lib-openshift.sh"
...
 
-set -eo nounset
+set -u
+
+TEST_LIST="\
+test_perl_s2i_sample_app
+test_perl_s2i_dancer_app
+test_perl_s2i_templates
+test_perl_imagestream
+"
```
phracek, line 24: What about adding here `set -e` so we are sure that we have a connection to OpenShift 4 cluster and before tests set `set +e`?

phracek (changes requested):

test/run-openshift-remote-cluster
zmiklank: IMHO we should treat this differently. E.g., calling exit explicitly, when the login is not successful. That way we at least have the debug message:
 ```oc status || false "It looks like oc is not properly logged in. Exiting the test suite" && TESTSUITE_RESULT=1; exit 1```
 instead of:
 ```oc status || false "It looks like oc is not properly logged in."```

phracek: It makes sense. Let's use this way.

zmiklank: Addressed in https://github.com/sclorg/container-common-scripts/pull/262. If that is merged I can use the function in this container.

zmiklank:
Added latest common/ and also added a line for exiting tests, when oc login unsuccessful.
[test-all]

zmiklank:
[test-all]

zmiklank:
[test-all]

phracek:
[test-all]

phracek:
[test-all]

phracek (approved):
The changes LGTM. Let's see whether tests are passing.

PR #249, opened 2022-06-23
Fix display-name in mod_fcgid variants

michal-josef-spacek:


pkubatrh:
lgtm

nathanweeks:
The builder image at registry.access.redhat.com still advertise, e.g., "Apache 2.4 with mod_perl/5.32":

https://catalog.redhat.com/software/containers/ubi9/perl-532/61a6115ec17162a20c1c6a67

Are those titles set in a different repository?

michal-josef-spacek:
@phracek @pkubatrh What about this name ^^^?

PR #269, opened 2023-03-29
Enable testing Perl 5.36

jplesnik:


phracek (approved):
Everything LGTM. THanks for this pull request and add new Perl version.

zmiklank:
Do not we want to add perl 5.36 also to build and push action?

phracek:
Adding new version to container mean, that `build-and-push-action` is modified as well.

zmiklank:
Yes, it will be failing until the version is added, but this is standard practice. The python container does it in this moment too.
It does not really matter though if you'll remember to add it later.

PR #270, opened 2023-03-29
Perl 5.36 support

jplesnik:
Add a new version Perl 5.36


5.36/Dockerfile.fedora
```diff
@@ -1,11 +1,11 @@
 FROM quay.io/fedora/s2i-base:36
```
phracek, line 1: Based on the https://src.fedoraproject.org/rpms/perl in Fedora 36 is perl-5.34. Can we please use `FROM quay.io/fedora/s2i-base:37`  where is perl-536?

phracek (changes requested):
Thanks for this PR! I hit only one issue during the review. Please address my comment.

zmiklank:
@jplesnik  Also could you please remove the merge commit, so we have cleaner git history?

phracek:
Let's try to test it.

[test]

5.36/README.md
```diff
@@ -0,0 +1,220 @@
...
+and web programming.
+
+This container image includes an cpanm utility, so users can use it to install Perl
+modules for their web applications. There is no guarantee for any specific CPAN module
+version, that is included in the image; those versions can be changed anytime.
+
+Usage in Openshift
+------------------
+
+In this example, we will assume that you are using the `ubi8/perl-536` image, available via `perl:5.36` imagestream tag in Openshift.
```
phracek, line 29: `ubi8/perl-536` should be change to `ubi8/perl-532` because there is no plan to have this version in RHEL8.

@pkubatrh WDYT? Am I right?

phracek (commented):
@pkubatrh Can you please help me with README.md issue? WDYT?

5.36/README.md
pkubatrh: I would move it to `ubi9/perl-536` instead. The README should be for the 5.36 version, I do not think it makes sense to have a different version of the image in there, even if it exists and the current version one does not.
I recommend moving to ubi9 instead of ubi8 for the sole reason that RHEL9 is more likely to receive new versions of Perl in future.

phracek:
[test-all]

phracek (approved):
All changes are addressed. Thank you for this pull request and the new perl version to Fedora containers land.

RHEL7 and RHEL OpenShift 4 failures are not affected by this pull request.

PR #288, opened 2023-11-09
Rewrite 5.34 to ${APP_ROOT}

michal-josef-spacek:


zmiklank:
[test-all]

zmiklank:
Code lgtm, let's wait for the tests results.

zmiklank:
I noticed just now, that we do not ship/publish 5.34 anymore. So we do not test this version either. 
I would merge this, as it is already written.

zmiklank (approved):

PR #298, opened 2024-04-04
Add check if new value is really present

phracek:
Add check if the new value is present in application and increase timeout for sure.


phracek:
[test]

michal-josef-spacek (approved):

PR #326, opened 2025-01-17
Support enabling perl-container on RHEL10

phracek:
This pull request only enables testing perl container on RHEL10 host

No code is changed only GitHub Action itself


phracek:
[test]

phracek:
Test passed. Let's get merge it.

PR #340, opened 2025-07-23
Deprecate version 5.32 for rhel8

PavolSloboda:
and update the README.md to reflect this change and suggest rhel9 instead

### DIFF
--- a/test/test_imagestreams_quickstart.py
+++ b/test/test_imagestreams_quickstart.py
@@ -25,6 +25,9 @@ if VERSION == "5.30-mod_fcgid":
 if VERSION == "5.26-mod_fcgid":
     VERSION = "5.26"
 
+if VERSION == "5.32" and OS == "rhel8":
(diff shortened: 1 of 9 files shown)


phracek:
[test-all]

test/test_imagestreams_quickstart.py
```diff
@@ -25,6 +25,9 @@
 if VERSION == "5.26-mod_fcgid":
     VERSION = "5.26"
 
+if VERSION == "5.32" and OS == "rhel8":
```
phracek, line 28: This should be mentioned at the beginning of function `test_perl_template_inside_cluster`. The rest is ok.

phracek (changes requested):
Only one more issue. The rest is OK. Thanks for the update.

phracek:
[test-all]

phracek (approved):
LGTM. Thanks

